PR #45, opened 2015-01-23
Fix AMD module loading

pedrocarrico:
Hi,

I am using AMD with an ember-cli app and it seems that JsDiff doesn't get loaded as a module unless you do a `define.amd` when you test for the `define` function.

I really don't know if this is correct pattern to do this but I've seen this behaviour in some other libraries and know it's loading correctly.

Cheers,

### DIFF
--- a/diff.js
+++ b/diff.js
@@ -532,7 +532,7 @@
   if (typeof module !== 'undefined' && module.exports) {
     module.exports = JsDiff;
   }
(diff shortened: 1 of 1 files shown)


kpdecker:
Released in 1.2.2

pedrocarrico:
Thanks for the fast release! :+1:

PR #47, opened 2015-02-09
Adding Diff Trimmed Lines

JamesGould123:
I extended the LineDiff.tokenize function to add the ability to ignore leading and trailing spaces when using TrimmedLineDiff, and added diffTrimmedLines to the public-facing functions. My changes should not affect any previous functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ or
 
     Returns a list of change objects (See below).
 
-* `JsDiff.TrimmedLineDiff(oldStr, newStr[, callback])` - diffs two blocks of text, comparing line by line, ignoring leading and trailing whitespace.
+* `JsDiff.diffTrimmedLines(oldStr, newStr[, callback])` - diffs two blocks of text, comparing line by line, ignoring leading and trailing whitespace.
 
     Returns a list of change objects (See below).
 
--- a/README.md
+++ b/README.md
@@ -33,6 +33,10 @@ or
 
     Returns a list of change objects (See below).
 
+* `JsDiff.TrimmedLineDiff(oldStr, newStr[, callback])` - diffs two blocks of text, comparing line by line, ignoring leading and trailing whitespace.
+
+    Returns a list of change objects (See below).
+
 * `JsDiff.diffSentences(oldStr, newStr[, callback])` - diffs two blocks of text, comparing sentence by sentence.
 
     Returns a list of change objects (See below).
@@ -166,4 +170,3 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 [![Bitdeli Badge](https://d2weczhvl823v0.cloudfront.net/kpdecker/jsdiff/trend.png)](https://bitdeli.com/free "Bitdeli Badge")
-
--- a/diff.js
+++ b/diff.js
@@ -270,20 +270,18 @@
             lastLineLastChar = lastLine ? lastLine[lastLine.length - 1] : '';
 
         // Merge lines that may contain windows new lines
-        if (line === '\n' && lastLine &&
-          (lastLineLastChar === '\r' || lastLineLastChar === '\n')) {
-          if(this.ignoreTrim || lastLineLastChar === '\n'){
+        if (line === '\n' && (lastLineLastChar === '\r' || lastLineLastChar === '\n')) {
+          if (this.ignoreTrim || lastLineLastChar === '\n'){
             //to avoid merging to \n\n, remove \n and add \r\n.
             retLines[retLines.length - 1] = retLines[retLines.length - 1].slice(0,-1) + '\r\n';
-          }
-          else{
+          } else {
             retLines[retLines.length - 1] += '\n';
           }
         } else if (line) {
-          if(this.ignoreTrim){
+          if (this.ignoreTrim) {
             line = line.trim();
             //add a newline unless this is the last line.
-            if(!(i + 1 === lines.length)){
+            if (i < lines.length - 1) {
               line += '\n';
             }
           }
@@ -469,7 +467,7 @@
             addEOFNL = false;
 
         for (var i = (diffstr[0][0]==='I'?4:0); i < diffstr.length; i++) {
-          if(diffstr[i][0] === '@') {
+          if (diffstr[i][0] === '@') {
             var meh = diffstr[i].split(/@@ -(\d+),(\d+) \+(\d+),(\d+) @@/);
             diff.unshift({
               start:meh[3],
@@ -478,17 +476,17 @@
               newlength:meh[4],
               newlines:[]
             });
-          } else if(diffstr[i][0] === '+') {
+          } else if (diffstr[i][0] === '+') {
             diff[0].newlines.push(diffstr[i].substr(1));
-          } else if(diffstr[i][0] === '-') {
+          } else if (diffstr[i][0] === '-') {
             diff[0].oldlines.push(diffstr[i].substr(1));
-          } else if(diffstr[i][0] === ' ') {
+          } else if (diffstr[i][0] === ' ') {
             diff[0].newlines.push(diffstr[i].substr(1));
             diff[0].oldlines.push(diffstr[i].substr(1));
-          } else if(diffstr[i][0] === '\\') {
+          } else if (diffstr[i][0] === '\\') {
             if (diffstr[i-1][0] === '+') {
               remEOFNL = true;
-            } else if(diffstr[i-1][0] === '-') {
+            } else if (diffstr[i-1][0] === '-') {
               addEOFNL = true;
             }
           }
@@ -498,7 +496,7 @@
         for (var i = diff.length - 1; i >= 0; i--) {
           var d = diff[i];
           for (var j = 0; j < d.oldlength; j++) {
-            if(str[d.start-1+j] !== d.oldlines[j]) {
+            if (str[d.start-1+j] !== d.oldlines[j]) {
               return false;
             }
           }
--- a/diff.js
+++ b/diff.js
@@ -115,8 +115,8 @@
             return done([{ value: newString, added: true }]);
           }
 
-          newString = this.tokenize(newString, !!self.ignoreTrim);
-          oldString = this.tokenize(oldString, !!self.ignoreTrim);
+          newString = this.tokenize(newString);
+          oldString = this.tokenize(oldString);
 
           var newLen = newString.length, oldLen = oldString.length;
           var maxEditLength = newLen + oldLen;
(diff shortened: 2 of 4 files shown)


JamesGould123:
I've realized that I did not add documentation in the README, I can do this as well once this pull request is merged.

kpdecker:
@JamesGould123 you can push additional commits to the branch with any changes that you'd like to make. Docs and tests are something I'd like to see before merging.

diff.js
```diff
@@ -114,8 +115,8 @@
             return done([{ value: newString, added: true }]);
           }
 
-          newString = this.tokenize(newString);
-          oldString = this.tokenize(oldString);
+          newString = this.tokenize(newString, !!self.ignoreTrim);
+          oldString = this.tokenize(oldString, !!self.ignoreTrim);
```
kpdecker: Rather than changing the tokenize API, can't we just read this flag below?

JamesGould123:
@kpdecker Sorry, I'm new to using Github, didn't know I could edit a pull request I've already sent, I'll add documentation and look into adding tests tonight.

diff.js
JamesGould123: Oh, certainly, thanks for the line note. I worked myself into a logical corner, and couldn't think of any other way to do that when I was coding, so I did this hesitantly. Looking at it again, the correct solution is quite clear.

kpdecker:
@JamesGould123 no worries. Welcome to github! :)

Just so you are aware, github does not send updates when the PR changes, so you should make a comment here saying that you've made the changes so I'll get a notification.

JamesGould123:
@kpdecker Thanks!

Okay, I've overcome my fear of npm, added tests, moved the edits I made to the tokenize API into the LineDiff.tokenize function, and added documentation.

README.md
```diff
@@ -33,6 +33,10 @@ or
 
     Returns a list of change objects (See below).
 
+* `JsDiff.TrimmedLineDiff(oldStr, newStr[, callback])` - diffs two blocks of text, comparing line by line, ignoring leading and trailing whitespace.
```
kpdecker: This should be `JsDiff.diffTrimmedLines`

test/diffTest.js
```diff
@@ -200,6 +200,36 @@ describe('#diffLines', function() {
...
+    diff.convertChangesToXML(diffResult).should.equal('line\n<ins>new value\n</ins><del>old value\n</del>line');
+  });
+  it('should the same lines in diff', function() {
+    var diffResult = diff.diffTrimmedLines(
+      'line\nvalue\nline',
+      'line\nvalue\nline');
+    diff.convertChangesToXML(diffResult).should.equal('line\nvalue\nline');
+  });
+
+  it('should ignore shorespace', function() {
```
kpdecker: Could you say something other than shorespace? This isn't really clear what the test is intending.

diff.js
```diff
@@ -256,18 +257,34 @@
...
           lines = value.split(/^/m);
-
       for(var i = 0; i < lines.length; i++) {
         var line = lines[i],
             lastLine = lines[i - 1];
 
         // Merge lines that may contain windows new lines
-        if (line === '\n' && lastLine && lastLine[lastLine.length - 1] === '\r') {
-          retLines[retLines.length - 1] += '\n';
+        if (line === '\n' && lastLine && lastLine[lastLine.length - 1]) {
```
kpdecker: The removal of the `'\r'` check seems like it will break something. Why was this removed?

diff.js
```diff
@@ -256,18 +257,34 @@
...
+          if(this.ignoreTrim){
+            retLines[retLines.length - 1] = retLines[retLines.length - 1].slice(0,-1) + '\r\n';
```
kpdecker: Why is this `\r\n` here but `\n` below?

README.md
JamesGould123: Sorry, my bad.

test/diffTest.js
JamesGould123: On line 188, in the tests for #diffLines:

<code>it('should handle shorespace', function() {</code>

I was reversing that test. Shorespace refers to the leading and trailing whitespace. I think both tests should be titled similarly, so they can be easily juxtaposed. Should I switch both to say "leading and trailing whitespace" instead of "shorespace"?

diff.js
JamesGould123: Huh, I'm surprised that didn't fail any of the tests. It passes all of the tests either way, but I'll re-add the "=== '\r'" because it should definitely be there. Maybe I should add a more verbose test for how it handles newlines.

diff.js
JamesGould123: string.trim() in JS removes the newline characters, which puts the entire diff on one line, so I added the '\n' character after trimming every line but the last line. However, then when it tried to merge empty lines using:

<code>retLines[retLines.length - 1] += '\n';</code>

It would end up with '\n\n' on the end of the line, instead of '\r\n' I fixed this by slicing and re-adding the '\r\n'.

JamesGould123:
Okay, I just pushed a commit fixing the documentation issue and extending the merge lines functionality and adding a comment explaining it.

JamesGould123:
@kpdecker Have you looked at the most recent changes? I've done everything you requested other than changing the test, which I explained and asked your opinion on above.

diff.js
```diff
@@ -256,18 +257,36 @@
...
-            lastLine = lines[i - 1];
+            lastLine = lines[i - 1],
+            lastLineLastChar = lastLine ? lastLine[lastLine.length - 1] : '';
 
         // Merge lines that may contain windows new lines
-        if (line === '\n' && lastLine && lastLine[lastLine.length - 1] === '\r') {
-          retLines[retLines.length - 1] += '\n';
+        if (line === '\n' && lastLine &&
+          (lastLineLastChar === '\r' || lastLineLastChar === '\n')) {
+          if(this.ignoreTrim || lastLineLastChar === '\n'){
```
kpdecker: Please match whitespace style. (Space after `if` and before `{`, no `else` on the same line)

diff.js
```diff
@@ -256,18 +257,36 @@
...
       for(var i = 0; i < lines.length; i++) {
         var line = lines[i],
-            lastLine = lines[i - 1];
+            lastLine = lines[i - 1],
+            lastLineLastChar = lastLine ? lastLine[lastLine.length - 1] : '';
 
         // Merge lines that may contain windows new lines
-        if (line === '\n' && lastLine && lastLine[lastLine.length - 1] === '\r') {
-          retLines[retLines.length - 1] += '\n';
+        if (line === '\n' && lastLine &&
```
kpdecker: This `lastLine` check isn't necessary since it was there to prevent a `undefined is not an object` error when reading from the array and that is now done when initializing the `lastLineLastChar` variable..

diff.js
```diff
@@ -256,18 +257,36 @@
...
+            retLines[retLines.length - 1] = retLines[retLines.length - 1].slice(0,-1) + '\r\n';
+          }
+          else{
+            retLines[retLines.length - 1] += '\n';
+          }
         } else if (line) {
+          if(this.ignoreTrim){
+            line = line.trim();
+            //add a newline unless this is the last line.
+            if(!(i + 1 === lines.length)){
```
kpdecker: nit pick: `if (i < lines.length - 1) {` save a few characters when minimized (or `i + 1`, the key thing is using the `<` operator rather than `!` and `===`)

test/diffTest.js
kpdecker: I think leading and trailing whitespace is a bit clearer. I didn't know what shorespace meant when I read it.

diff.js
```diff
@@ -256,18 +257,36 @@
...
+            lastLineLastChar = lastLine ? lastLine[lastLine.length - 1] : '';
 
         // Merge lines that may contain windows new lines
-        if (line === '\n' && lastLine && lastLine[lastLine.length - 1] === '\r') {
-          retLines[retLines.length - 1] += '\n';
+        if (line === '\n' && lastLine &&
+          (lastLineLastChar === '\r' || lastLineLastChar === '\n')) {
+          if(this.ignoreTrim || lastLineLastChar === '\n'){
+            //to avoid merging to \n\n, remove \n and add \r\n.
+            retLines[retLines.length - 1] = retLines[retLines.length - 1].slice(0,-1) + '\r\n';
```
kpdecker: I don't think we should use `\r\n` here. We are normalizing it to `\n` for all other lines so it's surprising to have it here and nowhere else. Instead, I would do nothing. I.e.

```
if (line === '\n' && lastLine && lastLine[lastLine.length - 1] === '\r') {
  if (!this.ignoreTrim) {
    retLines[retLines.length - 1] += '\n';
  }
}
```

We can do this because we know that the line will always have a `\n` at the end due to the `+=` operation below.

kpdecker:
A side note, I'm leaving for vacation shortly so my responses might take a bit longer, but I do want to get this merged!

diff.js
JamesGould123: I think you mean else should be on on the same line, it appears to be same line throughout the rest of the file, like this:
<pre>
if () {

} else {

}</pre>

test/diffTest.js
JamesGould123: Agreed, I'll change that.

diff.js
JamesGould123: Nice catch.

diff.js
JamesGould123: There are several places in applyPatch where these rules are not followed, I'll fix that as well.

diff.js
JamesGould123: This causes the tests I wrote to fail, which isn't necessarily a problem, but I followed the test set by the Line Diff tests, so I don't think it is exhibiting different behavior that that.

The final Line Diff test is:

<pre>
it('should handle windows line endings', function() {
    var diffResult = diff.diffLines(
      'line\r\nold value \r\nline',
      'line\r\nnew value\r\nline');
    diff.convertChangesToXML(diffResult).should.equal('line\r\n<ins>new value\r\n</ins><del>old value \r\n</del>line');
  });
</pre>

The final Trimmed Line Diff test is:

<pre>
  it('should handle windows line endings', function() {
    var diffResult = diff.diffTrimmedLines(
      'line\r\nold value \r\nline',
      'line\r\nnew value\r\nline');
    diff.convertChangesToXML(diffResult).should.equal('line\r\n<ins>new value\r\n</ins><del>old value\r\n</del>line');
  });
</pre>

Changing to your suggested code causes the second test to fail with the following value (no \r):

<code>'line\n<ins>new value\n</ins><del>old value\n</del
>line'</code>

This is the only thing I'm not sure about, everything else should be updated in my most recent commit.

JamesGould123:
@kpdecker, okay I've fixed everything except that last issue. Have a good time on your vacation!

kpdecker:
Released in 1.3.0

JamesGould123:
Thanks :smile:

PR #56, opened 2015-05-03
Two files patch

rgeissert:
When comparing strings that represent different versions, say 0.1.0/README and 0.1.1/README, it is nicer to include that metadata in the patch.
The proposed changes introduce createTwoFilesPatch, which allows it to take the old and new file names. createPatch calls the new function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,14 +53,19 @@ or
 
     Returns a list of change objects (See below).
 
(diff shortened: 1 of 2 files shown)


kpdecker:
Thanks!

kpdecker:
Released in 1.4.0

Issue #77, opened 2015-09-29
error throwing when apply patch to empty string


21paradox:
Here is the demo.
http://plnkr.co/edit/VSwrPtUuGSijx2JQBZNu?p=info

21paradox:
create pull request https://github.com/kpdecker/jsdiff/pull/78

PR #78, opened 2015-09-29
fix: error throwing when apply patch to empty string

21paradox:
This is related to issue https://github.com/kpdecker/jsdiff/issues/77

if previsousOperation doesn't exists, just continue.

also add two more test about this.


kpdecker:
Thanks!

kpdecker:
Released in 2.1.3

21paradox:
:+1: :+1:

Issue #82, opened 2015-10-13
Filenames not added to indexes


piranna:
When using `applyPatches()` with https://raw.githubusercontent.com/GregorR/musl-cross/master/patches/gcc-4.7.3-musl.diff I get the next result on `loadFile` callback:

```javascript
{ hunks: 
   [ { oldStart: 264,
       oldLines: 6,
       newStart: 264,
       newLines: 13,
       lines: [Object] },
     { oldStart: 272,
       oldLines: 6,
       newStart: 279,
       newLines: 9,
       lines: [Object] },
     { oldStart: 522,
       oldLines: 7,
       newStart: 522,
       newLines: 7,
       lines: [Object] },
     { oldStart: 625,
       oldLines: 6,
       newStart: 625,
       newLines: 9,
       lines: [Object] },
     { oldStart: 33,
       oldLines: 10,
       newStart: 33,
       newLines: 12,
       lines: [Object] },
     { oldStart: 54,
       oldLines: 18,
       newStart: 56,
       newLines: 21,
       lines: [Object] },
     { oldStart: 85,
       oldLines: 21,
       newStart: 90,
       newLines: 21,
       lines: [Object] },
     { oldStart: 108,
       oldLines: 3,
       newStart: 113,
       newLines: 74,
       lines: [Object] },
     { oldStart: 30,
       oldLines: 3,
       newStart: 30,
       newLines: 7,
       lines: [Object] },
     { oldStart: 184,
       oldLines: 6,
       newStart: 184,
       newLines: 7,
       lines: [Object] },
     { oldStart: 200,
       oldLines: 6,
       newStart: 201,
       newLines: 7,
       lines: [Object] },
     { oldStart: 215,
       oldLines: 6,
       newStart: 217,
       newLines: 7,
       lines: [Object] },
     { oldStart: 28,
       oldLines: 6,
       newStart: 28,
       newLines: 8,
       lines: [Object] },
     { oldStart: 47,
       oldLines: 28,
       newStart: 47,
       newLines: 13,
       lines: [Object] },
     { oldStart: 26736,
       oldLines: 6,
       newStart: 26736,
       newLines: 9,
       lines: [Object] },
     { oldStart: 26769,
       oldLines: 6,
       newStart: 26772,
       newLines: 7,
       lines: [Object] },
     { oldStart: 26851,
       oldLines: 6,
       newStart: 26855,
       newLines: 9,
       lines: [Object] },
     { oldStart: 4719,
[...]

PR #83, opened 2015-10-15
Allow patchs in diff format

piranna:
This fixes issue #82

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,9 +23,9 @@
   },
   "main": "./lib",
   "scripts": {
-    "test": "grunt"
+    "prepublish": "grunt build",
+    "test": "grunt test"
   },
-  "dependencies": {},
   "devDependencies": {
     "async": "^1.4.2",
     "babel": "^5.8.23",
@@ -60,6 +60,5 @@
     "semver": "^5.0.3",
     "webpack": "^1.12.2",
     "webpack-dev-server": "^1.12.0"
-  },
-  "optionalDependencies": {}
+  }
 }
--- a/src/patch/apply.js
+++ b/src/patch/apply.js
@@ -19,6 +19,7 @@ export function applyPatch(source, uniDiff, options = {}) {
 
       compareLine = options.compareLine || ((lineNumber, line, operation, patchContent) => line === patchContent),
       errorCount = 0,
+      offset = 0,
       fuzzFactor = options.fuzzFactor || 0,
 
       removeEOFNL,
@@ -46,18 +47,34 @@ export function applyPatch(source, uniDiff, options = {}) {
     return true;
   }
 
- // Sanity check the input string. Fail if we don't match.
+ // Search best fit offsets for each hunk based on the previous ones
  for (let i = 0; i < hunks.length; i++) {
-    let hunk = hunks[i];
+    let hunk = hunks[i],
+        minLine = 0,
+        toPos = offset + hunk.oldStart - 1;
 
-    if (!hunkFits(hunk, hunk.oldStart - 1)) {
-      return false;
+    for (let localOffset = 0; ; localOffset++) {
+      if (toPos - localOffset < minLine
+      || lines.length < toPos + localOffset + hunk.oldLines) {
+        return false;
+      }
+
+      if (hunkFits(hunk, toPos - localOffset)) {
+        hunk.offset = offset -= localOffset;
+        minLine = hunk.offset + hunk.oldStart + hunk.oldLines;
+        break;
+      }
+      if (hunkFits(hunk, toPos + localOffset)) {
+        hunk.offset = offset += localOffset;
+        minLine = hunk.offset + hunk.oldStart + hunk.oldLines;
+        break;
+      }
     }
   }
 
   for (let i = 0; i < hunks.length; i++) {
     let hunk = hunks[i],
-        toPos = hunk.newStart - 1;
+        toPos = hunk.offset + hunk.newStart - 1;
 
     for (let j = 0; j < hunk.lines.length; j++) {
       let line = hunk.lines[j],
--- a/src/patch/apply.js
+++ b/src/patch/apply.js
@@ -19,12 +19,16 @@ export function applyPatch(source, uniDiff, options = {}) {
 
       compareLine = options.compareLine || ((lineNumber, line, operation, patchContent) => line === patchContent),
       errorCount = 0,
-      offset = 0,
       fuzzFactor = options.fuzzFactor || 0,
+      minLine = 0,
+      offset = 0,
 
       removeEOFNL,
       addEOFNL;
 
+  /**
+   * Checks if the hunk exactly fits on the provided location
+   */
   function hunkFits(hunk, toPos) {
     for (let j = 0; j < hunk.lines.length; j++) {
       let line = hunk.lines[j],
@@ -47,42 +51,46 @@ export function applyPatch(source, uniDiff, options = {}) {
     return true;
   }
 
- // Search best fit offsets for each hunk based on the previous ones
- for (let i = 0; i < hunks.length; i++) {
+  // Search best fit offsets for each hunk based on the previous ones
+  for (let i = 0; i < hunks.length; i++) {
     let hunk = hunks[i],
         outOfLimits = 0,
         localOffset = 0,
-        minLine = 0,
         toPos = offset + hunk.oldStart - 1;
 
     for (;;) {
-      if (toPos + localOffset + hunk.oldLines <= lines.length) {
-        if (hunkFits(hunk, toPos + localOffset)) {
-          hunk.offset = offset += localOffset;
-          break;
-        }
-      } else {
+      // Check if trying to fit beyond text length, and if not, check it fits
+      // after offset location (or desired location on first iteration)
+      if (lines.length < toPos + localOffset + hunk.oldLines) {
         outOfLimits++;
+      } else if (hunkFits(hunk, toPos + localOffset)) {
+        hunk.offset = offset += localOffset;
+        break;
       }
 
-      // If trying to fit hunk outside both limits, return error
+      // If we tried to fit hunk before text beginning and beyond text lenght,
+      // then hunk can't be fit on the text so we raise an error
       if (outOfLimits === 2) {
         return false;
       }
 
+      // Reset checks of trying to fit outside text limits and increase offset
+      // of the current hunk relative to its desired location
       outOfLimits = 0;
       localOffset++;
 
-      if (minLine <= toPos - localOffset) {
-        if (hunkFits(hunk, toPos - localOffset)) {
-          hunk.offset = offset -= localOffset;
-          break;
-        }
-      } else {
+      // Check if trying to fit before text beginning, and if not, check it fits
+      // before offset location
+      if (toPos - localOffset < minLine) {
         outOfLimits++;
+      } else if (hunkFits(hunk, toPos - localOffset)) {
+        hunk.offset = offset -= localOffset;
+        break;
       }
     }
 
+    // Set lower text limit to end of the current hunk, so next ones don't try
+    // to fit over already patched text
     minLine = hunk.offset + hunk.oldStart + hunk.oldLines;
   }
 
--- a/src/patch/apply.js
+++ b/src/patch/apply.js
@@ -51,48 +51,67 @@ export function applyPatch(source, uniDiff, options = {}) {
     return true;
   }
 
+  function distanceIterator(toPos, minLine, maxLine) {
+    let wantForward = true,
+        backwardExhausted = false,
+        forwardExhausted = false,
+        localOffset = 1;
+
+    return function iterator() {
+      if (wantForward && !forwardExhausted) {
+        if (backwardExhausted) {
+          localOffset++;
+        } else {
+          wantForward = false;
+        }
+
+        // Check if trying to fit beyond text length, and if not, check it fits
+        // after offset location (or desired location on first iteration)
+        if (toPos + localOffset <= maxLine) {
+          return localOffset;
+        }
+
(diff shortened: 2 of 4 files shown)


kpdecker:
The `-r` portion feels pretty specific to the exact file mentioned in #82. Feels like we might want that to be more generic.

piranna:
> The -r portion feels pretty specific to the exact file mentioned in #82

Maybe, I'm not sure. I've looked for some info about that argument and seems to be related to [CVS revisions](http://march-hare.com/library/cvsnt/diff.html), that make sense. According to this text, there could appear zero, one or more times, so maybe I should change the RegExp to show this case... Other option is to ignore all the flags and get only the last fragment (the index path), what do you think?

kpdecker:
Lets ignore everything but the path component. Can you add a test case to cover this?

src/patch/apply.js
```diff
@@ -42,7 +39,30 @@ export function applyPatch(source, uniDiff, options = {}) {
...
+
+    return true;
+  }
+
+ // Sanity check the input string. Fail if we don't match.
+ for (let i = 0; i < hunks.length; i++) {
+    let hunk = hunks[i];
+
+    if (!hunkFits(hunk, hunk.oldStart - 1)) {
+      return false;
```
kpdecker: What behavior does this change that is necessary to do two iterations?

package.json
```diff
@@ -27,39 +27,39 @@
...
-    "mocha": "^2.2.4",
-    "phantomjs": "^1.9.17",
-    "semver": "^5.0.1",
-    "webpack": "^1.10.5",
-    "webpack-dev-server": "^1.10.1"
+    "mocha": "^2.3.3",
+    "phantomjs": "^1.9.18",
+    "semver": "^5.0.3",
+    "webpack": "^1.12.2",
+    "webpack-dev-server": "^1.12.0"
```
kpdecker: As long as the tests pass I'm fine taking version changes as is, but generally it's better to keep the changes in a PR down to just those needed to implement your feature. Makes it easier to review and less likely to be rejected for reasons that are tangential to your feature change.

src/patch/apply.js
piranna: This is WiP to allow to calculate the location offset where the hunks best fit. This is to fix issue #84

package.json
piranna: I agree. When I have it finished I can be able to split the commits in several feature branches so we can analize them separately.

piranna:
> Lets ignore everything but the path component. Can you add a test case to cover this?

I have reading some docs and thinking about this thing, and I believe we are taking a bad approach. We are considering it like the `Index:`, while that's only the filename when in fact the patch can have several revisions, so if we want to use it as an unique identifier, we should capture the full line. Other option would be not set the index at all and allow it to be undefined, problem is that currently this is not supported and all hunks of an undefined index are send at once and without the filenames, just the original problem of issue #82. Independtly of what we do with the `Index:` and the `diff -r` lines, maybe we should allow to undefined index and not send all the hunks at once... What do you think?

piranna:
It's said, what I means is that according to docs both the `Index:` and `diff` lines are superfluous headers that add metadata and can be ignored altogether, being `Index:` a somewhat unique file identifier, and the `diff` line just a command line that can generate the hunk. I think we shouldn't trust on them too much...

piranna:
Ok, offsets now are correctly calculated and NodeOS can be able to download all its build dependencies purely on Javascript :-) I'mgoing to write some tests to prevent regressions.

piranna:
Test for hunk offsets added.

package.json
kpdecker: I've cherry-picked these into master if you want to rebase to help simplify the PR.

src/patch/apply.js
```diff
@@ -42,7 +40,61 @@ export function applyPatch(source, uniDiff, options = {}) {
...
+
+ // Search best fit offsets for each hunk based on the previous ones
+ for (let i = 0; i < hunks.length; i++) {
+    let hunk = hunks[i],
+        outOfLimits = 0,
+        localOffset = 0,
+        minLine = 0,
+        toPos = offset + hunk.oldStart - 1;
+
+    for (;;) {
```
kpdecker: The logic here is really hard to follow and I worry that it's going to lead to confusion down the road. Rather than having side effects that terminate the loop in the middle of the block and the loop written as an infinite loop, I think it would be clearer to have the termination case expressed in the `for`/`while` statement.

src/patch/apply.js
```diff
@@ -42,7 +40,61 @@ export function applyPatch(source, uniDiff, options = {}) {
...
+
+      // If trying to fit hunk outside both limits, return error
+      if (outOfLimits === 2) {
+        return false;
+      }
+
+      outOfLimits = 0;
+      localOffset++;
+
+      if (minLine <= toPos - localOffset) {
```
kpdecker: Need more comments to describe what is going on in each phase here. It seems like this is scanning both backwards and forwards looking for a match, providing equal weighting to both directions (vs. scanning backwards, then forwards which could bias one outcome) but I'm really not clear. Ideally this would have a high level overview in comments and then links to any articles or similar resources that discuss the underlying algorithm (if this came from something like that)

package.json
piranna: :+1: I've rebased it.

src/patch/apply.js
piranna: It's an optimization, the alternative is to have the old code and use a conditional to check it, or do the check twice on the most common case (the hunk fits at the first try, when `localOffset` is zero). This has only moved the block to check the current or next line before the loop, so it's checked always in a loop rollback fashion. The same would be achieved by doing the check previously to enter the loop, but code is duplicated. I can add more comments if you don't see it clear.

src/patch/apply.js
piranna: You are right, that's the intended behaviour. The idea is that the fit should be near the original location, but we don't know if it's before or after it. Also, I'm taking in account the previous offset, since the line location will affect the next hunks too.

piranna:
I've improved the code documentation.

src/patch/apply.js
```diff
@@ -42,8 +44,66 @@ export function applyPatch(source, uniDiff, options = {}) {
...
+  }
+
+  // Search best fit offsets for each hunk based on the previous ones
+  for (let i = 0; i < hunks.length; i++) {
+    let hunk = hunks[i],
+        outOfLimits = 0,
+        localOffset = 0,
+        toPos = offset + hunk.oldStart - 1;
+
+    for (;;) {
```
kpdecker: After coming back to this a few days later, I'm still worried that I'm not going to be able to maintain this.

What if instead we extract out the iteration logic into an iterator? Something like:

```
function distanceIterator(target, minimum, maximum) {
  let wantForward = true,
      backwardExhausted = false,
      forwardExhausted = false,
      offset = 0;
  return function iterate() {
    // Continue iteration on the rear when the 
    if (forwardExhausted) {
      wantForward = false;
      offset++;
    }

    if (!wantForward) {
      // Check if we have a potential position behind us
      let candidate = target - offset;
      if (candidate > minimum) {
        wantForward = true;
        return candidate;
      } else {
        backwardExhausted = true;
      }
    }

    // Check if we have a potential position in front of us
    candidate = target + offset;
    if (candidate < maximium) {
      wantForward = false;
      offset++;
      return candidate;
    } else if (!backwardExhausted) {
      // We've exhausted the forward section, reenter to continue
      // iteration on the behind.
      forwardExhausted = true;
      return iterate();
    }
  };
}
```

Which can live in it's own file and we can unit test in isolation. This loop here then becomes something like:

```
for (let i = 0; i < hunks.length; i++) {
  let ...
      toPos = offest + hunk.oldStart + 1,
      maxLine = lines.length - hunk.oldLines,
      iterator = distanceIterator(toPos, minLine, maxLine),
      toCheck = iterator();

  while (toCheck !== null) {
    if (hunkFits(hunk, toCheck)) {
      hunk.offset = toCheck;
      break;
    }

    toCheck = iterator();
  }
}
```

(I have not actually tested or even run this code, so it might need some work to get it running)

I think separating general flip flopping iteration and checking the hunk itself into separate code will help with clarity as well as give us a nice reusable component for this iteration should we need it in other locations.

piranna: > I think separating general flip flopping iteration and checking the hunk itself into separate code will help with clarity as well as give us a nice reusable component for this iteration should we need it in other locations.

I find this code a little bit more complicated and less eficient, but I like the idea of separate the flip-flop logic from the main loop. I'll fetch your code and try to check it out :-)

piranna:
I've just splitted the logic to calculate candidates on an iterator. Maybe the code gets more readable, but I think it's more complicated and less performance, but at least it does the job.

src/patch/apply.js
```diff
@@ -42,7 +44,90 @@ export function applyPatch(source, uniDiff, options = {}) {
...
+  // Search best fit offsets for each hunk based on the previous ones
+  for (let i = 0; i < hunks.length; i++) {
+    let hunk = hunks[i],
+        maxLine = lines.length - hunk.oldLines,
+        localOffset = 0,
+        toPos = offset + hunk.oldStart - 1;
+
+    let iterator = distanceIterator(toPos, minLine, maxLine);
+
+    for (; localOffset != undefined; localOffset = iterator()) {
```
kpdecker: You need `!==` here or `!= null` otherwise zero will fail. We should have a test case for that.

src/patch/apply.js
```diff
@@ -42,7 +44,90 @@ export function applyPatch(source, uniDiff, options = {}) {
...
           }
         }
+        toPos++;
       }
+    }
+
+    return true;
+  }
+
+  function distanceIterator(toPos, minLine, maxLine) {
```
kpdecker: This should probably be a separate module to help make things clearer.

src/patch/apply.js
piranna: > You need !== here or != null otherwise zero will fail.

```Javascript
[piranna@Mabuk:~/Proyectos/NodeOS/node_modules]
 (vagga) > node
> 0 != undefined
true
> 
```

I think it's not needed... :-) But I could agree on the `!== undefined`, in any case is the only thing it will return... I left it this way to be more generic.


> We should have a test case for that.

A test case for what, exactly?

kpdecker:
Thanks for making the changes.

Regarding complicated, by breaking out the components, it should be easier to think about and diagnose what is going. Not always the case, but I've been burned by a few patches in this repo that I didn't fully understand that caused me great headaches down the road.

Regarding performance this is still the same O, something like `O(m*n)` while it would need to be tested, I presume that this would perform at a similar level. The repeated substr, etc in the `hunkFits` logic feels like it will have more if an impact than a few function calls.

src/patch/apply.js
kpdecker: Perhaps it's been too long that I've been using `!= null`. Generally prefer `!= null` for "anything other than null and undefined" and "!== undefined" for "just looking for undefined". (I actually thought that was in the linter rules, but I guess not).

Little bit of a nit picky change but would be nice to have so it doesn't cause confusion down the road.

Test case: To make sure that it can patch at an offset of zero.

piranna:
> Regarding complicated, by breaking out the components, it should be easier to think about and diagnose what is going. Not always the case, but I've been burned by a few patches in this repo that I didn't fully understand that caused me great headaches down the road.

Sometimes overengineering promotes this (at least in my case), so I try to keep things as simple and minimal as possible, so it's also less error-prone.


> Regarding performance this is still the same O, something like O(m*n) while it would need to be tested, I presume that this would perform at a similar level.

It's the same O, but implementation is more verbose so it will be slower, that's what I said about performance.

package.json
```diff
@@ -60,6 +60,5 @@
     "semver": "^5.0.3",
     "webpack": "^1.12.2",
     "webpack-dev-server": "^1.12.0"
-  },
-  "optionalDependencies": {}
+  }
```
kpdecker: Please keep the changes focused on this PR.

piranna:
> Little bit of a nit picky change but would be nice to have so it doesn't cause confusion down the road.

Done.


> Test case: To make sure that it can patch at an offset of zero.

That's the trivial case, when offset is zero it fitted at first try, so it didn't need to exec the `iterator()` at all... It's already tested by the other test cases.

piranna:
I was able to put `distanceIterator()` as a module-level private function without problems. I tried to do the same with `hunkFits()` but has some secondary effects, so better left it as is.

So, to summarize, what changes need to be done (if any) so this pull-request can land on upstream? I need this to be available as soon as possible for so NodeOS (my bachelor thesis) can work... :-/

kpdecker:
Merging this now and I'll take care of the rest of the cleanup.

kpdecker:
This was merged, but the combination of rebase and master on master merge from the CLI commands caused github to not detect this. Closing this out, but it is merged.

kpdecker:
048e7f7 is what I was intending. Doing this and explicit functionality tests in the micro scale helps dramatically with code quality. Best of luck with your bachelors.

piranna:
> This was merged, but the combination of rebase and master on master merge from the CLI commands caused github to not detect this

Yes, this happened to me before. Sometimes it detect it, though. Don't worry :-)


> 048e7f7 is what I was intending. Doing this and explicit functionality tests in the micro scale helps dramatically with code quality

Ok, I see. I'm not too much to split such code to independent files (I prefer to have it near to where it's used, or at least as global private function to reduce memory usage) except if it has a clear own entity (so maybe I would put it on its own module) or if it's being used from several places.

I like the description that you put there, simple and clear. I didn't think about it, but would be a good candidate for a simple test :-)


> Best of luck with your bachelors

Thank you! :-) You can see how it goes on at https://github.com/NodeOS/NodeOS :-)

PR #86, opened 2015-10-29
Generate library on 'npm install' and 'npm publish' automatically

piranna:
This allow to build automatically the `lib` folder when code is downloaded from GitHub and just before to upload it to npm. Also it removes empty dependencies objects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,9 +23,9 @@
   },
   "main": "./lib",
   "scripts": {
-    "test": "grunt"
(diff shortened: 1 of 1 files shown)


kpdecker:
This is highly error prone and requires a lot of extra dependencies which is why I removed it from the other PR. Not the route we want to go.

piranna:
It good to have it on development environments, but it's ok if you don't want it.

PR #87, opened 2015-10-29
Split diffs using file headers

piranna:
This follow better the structure of the "Unified Diff" format instead of depending on optional non-standard metadata like "Index:" or "diff" lines. This metadata is still parsed and used, though, but now it's possible to include other more easily. The main advantage of this is that now it's possible to use `applyPatches()` also with single anonimous diffs, that allow to use [download-manager](https://github.com/piranna/download-manager) to download and patch all the source code of NodeOS without corner cases.

### DIFF
--- a/src/patch/parse.js
+++ b/src/patch/parse.js
@@ -78,22 +78,26 @@ export function parsePatch(uniDiff, options = {}) {
 
     let addCount = 0,
         removeCount = 0;
+
     for (; i < diffstr.length; i++) {
-      let operation = diffstr[i][0];
+      let line = diffstr[i],
+          operation = line[0];
 
-      if (operation === '+' || operation === '-' || operation === ' ' || operation === '\\') {
-        hunk.lines.push(diffstr[i]);
-
-        if (operation === '+') {
-          addCount++;
-        } else if (operation === '-') {
-          removeCount++;
-        } else if (operation === ' ') {
-          addCount++;
-          removeCount++;
-        }
-      } else {
+      // Check begin of line is NOT from a hunk, end parsing
+      if (operation !== '+'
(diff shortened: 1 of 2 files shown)


kpdecker:
This is too much to review. As I mentioned on your past PR, you need to focus on individual changes. Broad scale renames that appear to be simply stylistic, etc mask the true changes. Please file a separate PR with just the functional changes.

I've noticed that you are committing directly to your fork's master, which can compound this problem. When doing pull requests you're better off creating a topic branch with the isolated changes and you can then merge those changes into your own master and use them directly from there while waiting for changes to land upstream. Ex: https://github.com/gotwarlost/istanbul/pull/422 is on the source-map-errors branch but I link to the master branch which could have multiple different active PRs merged in without impacting one another.

piranna:
> I've noticed that you are committing directly to your fork's master,

I'm doing it this way because some changes goes on top of another, for example the ones of the rename of `index` to `diff`, because in fact we are parsing a `diff` :-) Now that I have it working on my machine, I'll try to cherry-pick and refactor it in several pull-requests so it will be easier to review (although the content of each one will be mostly the same of each independent commit...).

kpdecker:
To be clear, I'm not really interested in stylistic refactorings so don't burn too much time there. "diff" vs. "index" is somewhat subjective and changing doesn't warrant the potential for errors being introduced in my eyes.

piranna:
https://github.com/kpdecker/jsdiff/pull/88 should have the bare minimum changes to allow this feature, with some minor improvements on the regexp. I'll left the bigger changes for upcoming pull-requests, since this change is the one is the only really needed by download-manager.

PR #88, opened 2015-10-29
Split diffs based on file headers instead of 'Index:' metadata

piranna:


src/patch/parse.js
```diff
@@ -7,41 +7,38 @@ export function parsePatch(uniDiff, options = {}) {
     let index = {};
     list.push(index);
 
-    // Ignore any leading junk
-    while (i < diffstr.length) {
-      if (/^(Index:|diff -r|@@)/.test(diffstr[i])) {
+    // Parse diff metadata
+    for (; i < diffstr.length; i++) {
```
kpdecker: Not going to accept purely stylistic changes for fear of the noise impact. Show me the substance not your own style. When working in the OSS world you have to remember that you aren't coding for yourself unless it's your own project.

piranna: This is not purely stylistic, this is to mimic the same style done on [line 78](https://github.com/kpdecker/jsdiff/pull/88/files#diff-a74b5a56ea1528866a8b97a54a3dfd6bR78) on this same file to have an uniform style on both places.

kpdecker: "This is to mimic style" means it's stylistic. I'm glad to discuss stylistic changes in another PR if you want to spend your time on that but I don't accept them for functional changes.

kpdecker: To that end style is anything that has the same effective functionality but just has different actual code.

The reason for this is that the functional changes are the important part and they should be focused on in isolation.

piranna: Ok, removed style changes.

kpdecker:
Still has some noise from the `line` vs. `diffstr[i]` changes.

In the other PR you reference unified diff format, did you base this off of a spec? Could you point to where you got this from? It's still not clear to me from the code what the changes are here other than making things a bit less accepting by adding additional regex restrictions.

I think the best way for you to express this is to write a test case that shows what would have failed before and how it doesn't now.

piranna:
> In the other PR you reference unified diff format, did you base this off of a spec? Could you point to where you got this from?

http://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html

As you can see there, the __only__ important lines are the ones of the file header (+++ and ---), the beginning of the hunks (@@) and the description of the added and removed ones (+ and -). Any of the other lines (metadata and context) are purely optional.


> I think the best way for you to express this is to write a test case that shows what would have failed before and how it doesn't now.

Trying to use a diff file with no Index line doesn't work with `applyPatches()`, this changes are not based on it and instead only on the file header lines, so it's more compliant with the standard and allow them to work.

piranna:
Any update on this?

src/patch/parse.js
```diff
@@ -51,7 +50,7 @@ export function parsePatch(uniDiff, options = {}) {
   // Parses the --- and +++ headers, if none are found, no lines
   // are consumed.
   function parseFileHeader(index) {
-    let fileHeader = (/^(\-\-\-|\+\+\+)\s(\S+)\s?(.*)/.exec(diffstr[i]));
+    let fileHeader = /^(\-\-\-|\+\+\+)\s+(\S+)\s?(.+)/.exec(diffstr[i]);
```
kpdecker: Your parenthesis change here introduces a linter error.

src/patch/parse.js
```diff
@@ -7,41 +7,40 @@ export function parsePatch(uniDiff, options = {}) {
...
     list.push(index);
 
-    // Ignore any leading junk
+    // Parse diff metadata
     while (i < diffstr.length) {
-      if (/^(Index:|diff -r|@@)/.test(diffstr[i])) {
+      let line = diffstr[i];
+
+      // File header found, end parsing diff metadata
+      if (/^(\-\-\-|\+\+\+|@@)/.test(line)) {
```
kpdecker: `@@` isn't in the spec is it? This seems to be a change the previous behavior.

test/patch/parse.js
```diff
@@ -248,8 +248,8 @@ Index: test2
     });
     it('should throw on invalid input in strict mode', function() {
       expect(function() {
-        parsePatch('Index: foo\nfoo', {strict: true});
-      }).to['throw'](/Unknown line 2 "foo"/);
+        parsePatch('Index: foo\n+++', {strict: true});
+      }).to['throw'](/Unknown line 2 "\+\+\+"/);
```
kpdecker: Please add positive test cases.

src/patch/parse.js
```diff
@@ -7,41 +7,40 @@ export function parsePatch(uniDiff, options = {}) {
...
     index.hunks = [];
 
     while (i < diffstr.length) {
-      if (/^(Index:|diff -r)/.test(diffstr[i])) {
+      let line = diffstr[i];
+
+      if (/^(Index:|diff|\-\-\-|\+\+\+)\s/.test(line)) {
         break;
-      } else if (/^@@/.test(diffstr[i])) {
+      } else if (/^@@\s/.test(line)) {
```
kpdecker: More restrictive than before. Why?

src/patch/parse.js
piranna: You are right, I missed it. Done.

src/patch/parse.js
piranna: I left it because it was set before. We could remove it if you want, but it doesn't hurt and would allow to accept patches without the file headers (I was thinking about propose it in other pull-request).

src/patch/parse.js
piranna: To prevent the case where the line start with @@whatever, that in strict mode it should throw an error instead of think it's a new hunk.

piranna:
Changes uploaded

src/patch/parse.js
kpdecker: Strict mode?

kpdecker: Relax both this and the changes down below. This is not a spec we are dealing with but a bunch of random projects that are hobbled together based on arbitrary documentation. We should follow Postel's law here.

test/patch/parse.js
```diff
@@ -248,8 +248,8 @@ Index: test2
     });
     it('should throw on invalid input in strict mode', function() {
       expect(function() {
-        parsePatch('Index: foo\nfoo', {strict: true});
-      }).to['throw'](/Unknown line 2 "foo"/);
+        parsePatch('Index: foo\n+++ bar\nblah', {strict: true});
+      }).to['throw'](/Unknown line 3 "blah"/);
```
kpdecker: When I said positive case, I meant create a new test case that would be failing without the changes above. Basically I'd like to see a TDD artifact here so we both have a example of what this change fixes if someone looks back on this in the future. This also help protect yourself as it ensures that your use case doesn't regress in the future.

src/patch/parse.js
piranna: I think we should follow the spec and the projects bad using it should fix themselves because being less restrictive here would lead to errors otherwise, but since we already are accepting hunks without file headers (that's not supported by the standard, but there's no technical issues for accept it) and there's other projects using the library that would break, I'll remove the `\s`.

test/patch/parse.js
piranna: Ok, I missed the point. I've added a test to check `applyPatches()` work with diffs without `Index` metadata, using `oldFileName` and `newFileName` fields, that's just my use case.

kpdecker:
I think we're close to there, but something confused github and while I see your last commit in the commit list, I don't see the changes in the total change list for the PR. Haven't really seen that before and not sure what happened, but maybe if you squash the changes that will fix the issue.

piranna:
I've just review it at https://github.com/kpdecker/jsdiff/pull/88/files and changes have been applied, could you be able to take a look again? I'll be connected all afternoon if there's some problem.

piranna:
Is there any other thing that needs to be fixed so this get merged and published? If it's a problem with GitHub merge you could pull the changes from my repo, publish from there and close this pull-request by hand... I need to return my laptop by this weekend and would like to get this finished by then

piranna:
Thank you, and sorry for my insistence. Could you be able to estimate when this will get published on npm?

kpdecker:
Released in 2.2.1

piranna:
I've just checked it from npm and works correctly, thank you.

PR #90, opened 2015-11-12
Use `switch` statement instead of concatenated `if`

piranna:
This is just a code clean-up, but I find it easier to read and understand.


kpdecker:
Thanks, but not really a fan of this style.

piranna:
Don't matter, everybody has its own styles :-)

PR #115, opened 2016-05-28
Fixed grammar in README

krizalys:
Fixed grammar in README:

`These objects are consist of the following fields:` => ` These objects consist of the following fields:`

My editor also removed trailing spaces, which is probably another good thing to do.


kpdecker:
Thanks!

kpdecker:
Released in 2.2.3

PR #118, opened 2016-05-31
Add a fix for applying 0-length destination patches

### LINKED ISSUE #116: applyPatch can't handle clearing the file
If a patch is supposed to completely erase a file, `applyPatch` will instead keep the first line.

    $ cat testfile 
    This
    is
    a
    file,
    yo
    $ cat > testfile          # empty the file out
    $ cat testfile
    $ git diff > diff.patch   # create unified patch w/ git
    $ cat diff.patch 
    diff --git a/testfile b/testfile
    index c9d61c6..e69de29 100644
    --- a/testfile
    +++ b/testfile
    @@ -1,5 +0,0 @@
    -This
    -is
    -a
    -file,
    -yo
    $ git checkout -- .       # restore testfile
    $ cat testfile 
    This
    is
    a
    file,
    yo
    $ patch -p1 < diff.patch  # test with 'patch'
    patching file testfile
    $ cat testfile            # yep, worked
    $ git checkout -- .       # restore testfile again
    $ cat patch.js
    var fs = require('fs');
    var diff = require('diff');
    
    console.log(
      diff.applyPatch(
        fs.readFileSync('testfile', 'utf8'),
        fs.readFileSync('diff.patch', 'utf8')
      )
    );
    
    $ node patch.js           # this should be empty
    This

chaaz:
If the destination lines length is 0, then actual editing should start one line later (since, per the spec, zero-length blocks are indexed at the line **before** any changes). This closes #116.

### DIFF
--- a/src/patch/apply.js
+++ b/src/patch/apply.js
@@ -81,6 +81,7 @@ export function applyPatch(source, uniDiff, options = {}) {
   for (let i = 0; i < hunks.length; i++) {
     let hunk = hunks[i],
         toPos = hunk.offset + hunk.newStart - 1;
+    if (hunk.newLines == 0) { toPos++; }
 
     for (let j = 0; j < hunk.lines.length; j++) {
       let line = hunk.lines[j],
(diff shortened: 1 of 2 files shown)


kpdecker:
Thanks!

kpdecker:
Released in 2.2.3

PR #126, opened 2016-08-05
Default oldStart and newStart values to 1 if NaN

kdz:
In diffs with conflicting changes, set old and new starting values to 1 (or other default value) if NaN. Allows for easier json parsing, as NaN is not recognized by json.


kpdecker:
This looks like it causes a build failure. I suspect that you want an explicit `isNaN` check as 0 is a valid value here (I think, it's been awhile since I've looked at this code).

PR #130, opened 2016-08-18
Add callback argument to applyPatches `patched` option

### LINKED ISSUE #129: Make `applyPatches` patched callback async
On `applyPatches()` functions the `loadFile()` option accept a callback that must be called with the content of the file to be patched, but the `patched()` option don't, forcing to implement it as a sync function. Ironically, [just in the next line](https://github.com/kpdecker/jsdiff/blob/4a899c0afc95228d20b6b743ce9ea45e4233d60e/src/patch/apply.js#L140-L142) there's a `setTimeout(processIndex, 0)` to process the next index!

What I propose is to add a callback argument to `patched()` function so it can both notify when it has finished and can start processing the next index, and also if there was an error while writting the patch. A more advanced option would be to use `async.js` for that, that's already a dependency and would make the code a bit cleaner.

piranna:
This fixes #129

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ or
     This method will iterate over the contents of the patch and apply to data provided through callbacks. The general flow for each patch index is:
 
     - `options.loadFile(index, callback)` is called. The caller should then load the contents of the file and then pass that to the `callback(err, data)` callback. Passing an `err` will terminate further patch execution.
-    - `options.patched(index, content)` is called once the patch has been applied. `content` will be the return value from `applyPatch`.
+    - `options.patched(index, content, callback)` is called once the patch has been applied. `content` will be the return value from `applyPatch`. When it's ready, the caller should call `callback(err)` callback. Passing an `err` will terminate further patch execution.
 
     Once all patches have been applied or an error occurs, the `options.complete(err)` callback is made.
 
--- a/src/patch/apply.js
+++ b/src/patch/apply.js
@@ -137,9 +137,13 @@ export function applyPatches(uniDiff, options) {
       }
 
       let updatedContent = applyPatch(data, index, options);
-      options.patched(index, updatedContent);
+      options.patched(index, updatedContent, function(err) {
+        if (err) {
+          return options.complete(err);
+        }
 
-      setTimeout(processIndex, 0);
+        processIndex();
+      });
     });
   }
   processIndex();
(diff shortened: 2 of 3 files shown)


piranna:
> This one looks good !

Thanks! ^^

kpdecker:
It's generally bad form to combine multiple features into one pull request. Please back out #131 from this as I don't want that in here.

WRT this exact feature, I'm worried about making this a breaking change. I'm not sure if it would be better to do something like `patchedAsync` or similar. Can you walk me through the use case that requires the patched callback to be async?

piranna:
> It's generally bad form to combine multiple features into one pull request. Please back out #131 from this as I don't want that in here.

I added it here since they were somewhat related. I can be able to remove it #131 is not open for discussion.


> WRT this exact feature, I'm worried about making this a breaking change. I'm not sure if it would be better to do something like patchedAsync or similar.

Another option would be to check the number of arguments of the function to see if it has defined a callback (`options.patched.length === 3`) and if so use it async, and otherwise sync.


> Can you walk me through the use case that requires the patched callback to be async?

On [download-manager](https://github.com/piranna/download-manager) I'm using it to apply patches to Linux kernel, GCC and some other tools for NodeOS and [write them back to the filesystem](https://github.com/piranna/download-manager/blob/bf6bcf4b675157f817f24f0345c57422622cbf81/index.js#L57-L68) before start processing or compiling them, so I need both to be sure they have been fully written before using them and to capture any possible error when writting to manage it cleanly. I could be able to write them synchronously to the filesystem and using a try-catch, but all the module is fully async and this would be like an alien, and also it would became a bottleneck and decrease performance (I'm downloading, decompressing, patching and writting the files on the fly...).

kpdecker:
Ok, I think that makes sense re: async w/ async. I don't want to go the route of checking the args length because that can also have unexpected conseqences, but I'm fine merging this and making it a major release after the changes for #131 are removed.

piranna:
> I don't want to go the route of checking the args length because that can also have unexpected conseqences

I agree, it's good to know that kind of magic but it's not necesary to use it often...


> but I'm fine merging this and making it a major release after the changes for #131 are removed

Changes for issue #131 reverded.

kpdecker:
Looks like even with the revert, there is still coverage missing. Could you make the tests green?

piranna:
> Looks like even with the revert, there is still coverage missing. Could you make the tests green?

Done, I wanted to check validity of the pull-request before.

piranna:
Now tests coverage is passing on CI server

kpdecker:
Released in 3.0.0

piranna:
Thank you :-)

PR #133, opened 2016-08-31
Default oldStart and newStart to -1 if NaN (appears in conflict diff)

kdz:
In diffs with conflicting changes, replace the NaN old and new
start values with -1. Allows easier JSON parsing, as NaN is not
recognized by JSON.


kdz:
The failure occurs in a branches test. Could you offer any advice on how to access the test and fix? 
My branch is at V3.0.0 which I noticed breaks at 224, this commit breaks at 225.
Thanks!


untitled drawing

kpdecker:
Add a test case that hits the isNaN positive code path. I.e. write a test case for the bug you're trying to fix.

PR #144, opened 2016-10-12
Split file using all possible line delimiter instead of hard-coded "/n" and join lines back using the original delimiters

soulbeing:
Split file using all possible line delimiter instead of hard-coded "/n" and join lines back using the original delimiters. 
let me know your feed back on this please.

### DIFF
--- a/src/patch/apply.js
+++ b/src/patch/apply.js
@@ -15,7 +15,8 @@ export function applyPatch(source, uniDiff, options = {}) {
   }
 
   // Apply the diff to the input
-  let lines = source.split('\n'),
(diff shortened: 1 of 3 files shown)


soulbeing:
/home/travis/build/kpdecker/jsdiff/node_modules/istanbul/lib/cli.js:38
        throw ex; // turn it into an uncaught exception
        ^
ERROR: Coverage for branches (99.8%) does not meet global threshold (100%)

this is the only error left, I don't think I can fix this.

kpdecker:
Released in 3.1.0

ExplodingCabbage:
@soulbeing do you remember what this was supposed to fix? The overall behaviour change effected by this PR is non-obvious and there are no tests demonstrating it.

PR #146, opened 2016-10-20
JsDiff.diffArrays to compare arrays

wvanderdeijl:
adds support for comparing two arrays and finding the difference. This comes in very handy in my current project. 
If you feel this is beyond the boundaries of the primary focus of jsdiff which is text comparison, I could trim the pull request to just having the .join method move to the base class so we can create the diffArrays in our own project. Having the .join('') in several methods of base prevents this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,10 @@ or
 
     Returns a list of change objects (See below).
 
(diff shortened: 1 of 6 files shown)


kpdecker:
Sorry for the delay. Generally looks good. Going to try to make some time soon to land this and get a new release out.

kpdecker:
Released in 3.1.0

PR #153, opened 2016-12-02
feat(structuredPatch): Pass options to diffLines

Kiougar:
Simply allow the user to use diffLines options with the structuredPatch function


kpdecker:
Released in 3.2.0

PR #156, opened 2016-12-08
Add `undefinedReplacement` option to `diffJson`

### LINKED ISSUE #155: Option to take in account undefined values in `diffJson`
I'm using `jsdiff` to show a diff if two objects are not equal. 
 
```js
const { diffJson } = require('diff');
const isEqual = require('lodash.isequal');

check({ x: 1, y: undefined }, { x: 1 });

function check(expected, actual) {
  if (!isEqual(expected, actual)) { //=> false
    console.log(diffJson(expected, actual)); //=> [ { value: '{\n  "x": 1\n}', count: 3 } ]
  }
}
```

Now it's ignoring `undefined` value due to default `JSON.stringify` behaviour. 
I propose to add an option to `diffJson` method to show changes. The simplest way I found is to add a custom `replacer` function (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) in `JSON.stringify` (see https://github.com/ewnd9/jsdiff/commit/c6daea71b07e660870a6e81b58b7d487fac1cb52) 

I would love to make a PR if the change is welcomed

ewnd9:
Fixes #155

Please, let me know if I need to change styling / rename `undefinedReplacement` to something more concise / anything else.

### DIFF
--- a/src/diff/json.js
+++ b/src/diff/json.js
@@ -11,14 +11,21 @@ jsonDiff.useLongestToken = true;
 
 jsonDiff.tokenize = lineDiff.tokenize;
 jsonDiff.castInput = function(value) {
-  return typeof value === 'string' ? value : JSON.stringify(canonicalize(value), undefined, '  ');
+  const {undefinedReplacement} = this.options;
+
+  return typeof value === 'string' ? value : JSON.stringify(canonicalize(value), function(k, v) {
+    if (typeof v === 'undefined') {
+      return undefinedReplacement;
+    }
+
+    return v;
+  }, '  ');
 };
 jsonDiff.equals = function(left, right) {
   return Diff.prototype.equals(left.replace(/,([\r\n])/g, '$1'), right.replace(/,([\r\n])/g, '$1'));
 };
 
-export function diffJson(oldObj, newObj, callback) { return jsonDiff.diff(oldObj, newObj, callback); }
-
+export function diffJson(oldObj, newObj, options) { return jsonDiff.diff(oldObj, newObj, options); }
 
 // This function handles the presence of circular references by bailing out when encountering an
 // object that is already on the "stack" of items being processed.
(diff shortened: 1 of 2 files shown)


src/diff/json.js
```diff
@@ -11,14 +11,33 @@ jsonDiff.useLongestToken = true;
...
+  if (typeof options === 'function') {
+    callback = options;
+    options = {};
+  }
 
+  if (typeof options === 'undefined') {
+    options = {};
+  }
+
+  jsonDiff.undefinedReplacement = options.undefinedReplacement;
```
ewnd9: I had to assign `undefinedReplacement` directly to `jsonDiff` because `this.options` is overriden each time at the `base#diff` method (https://github.com/kpdecker/jsdiff/blob/master/src/diff/base.js#L10).

test/diff/json.js
```diff
@@ -66,6 +66,27 @@ describe('diff/json', function() {
...
+        {a: 123, b: 456},
+        {undefinedReplacement: null}
+      )).to.eql([
+        { count: 3, value: '{\n  "a": 123,\n  "b": 456,\n' },
+        { count: 1, value: '  "c": null\n', added: undefined, removed: true },
+        { count: 1, value: '}' }
+      ]);
+    });
+
+    it('should handle callback', function(done) {
```
ewnd9: I had to add this to keep coverage at 100% which was dropped due to code below

```js
if (typeof options === 'function') {
  callback = options;
  options = {};
}
```

src/diff/json.js
kpdecker: Humm. This is not ideal as it causes side effects in the global space. Did you try to pass options through?

ewnd9:
My apologies for distraction, it's indeed working with just options.

If I recall correctly, I was trying to keep signature of `jsonDiff.diff(oldObj, newObj, callback)` but now I see it is already handled at https://github.com/kpdecker/jsdiff/blob/master/src/diff/base.js#L6-L9 

---

I've pushed a new commit with changes

ewnd9:
@kpdecker Could you release a new version on npm, please?

kpdecker:
Released in 3.2.0

PR #185, opened 2017-07-01
Add browser field pointing to dist/diff.js

porsager:
Having the browser field in package.json point to the bundle will provide better support for unpkg.com and bundlers wanting to use the dist file etc.


porsager:
Wow. How did that single apostrophe sneak in there. Sorry about that, and thanks for adding it properly 😉

Issue #183, opened 2017-06-13
Feature request: ability to specify a custom equality checker for `diffArrays`


tashburn:
I'd like to be able to compare arrays of objects with deep equality checking.

PR #187, opened 2017-08-03
comparator for custom equality checks

schornio:
Extended equal-function to support a custom comparator. Useful for `diffArray`.

Solves #183

### DIFF
--- a/src/diff/base.js
+++ b/src/diff/base.js
@@ -144,8 +144,15 @@ Diff.prototype = {
   },
 
   equals(left, right) {
-    return left === right
-      || (this.options.ignoreCase && left.toLowerCase() === right.toLowerCase());
+    if (this.options.comparator) {
+      if (Array.isArray(right)) {
+        right = right[0];
+      }
+      return this.options.comparator(left, right);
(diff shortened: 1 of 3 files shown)


src/diff/base.js
```diff
@@ -144,8 +144,15 @@ Diff.prototype = {
   },
 
   equals(left, right) {
-    return left === right
-      || (this.options.ignoreCase && left.toLowerCase() === right.toLowerCase());
+    if (this.options.comparator) {
+      if (Array.isArray(right)) {
+        right = right[0];
+      }
```
kpdecker: What is this array lookup doing? Is there a case where arrays are passed in here?

schornio: Yes there is some edge case - maybe because of how this lib uses `tokenize()` and `join()`

    arrayDiff.tokenize = arrayDiff.join = function(value) {
      return value.slice();
    };

schornio:
Any chance this gets merged?

kpdecker:
Merged, but rebased and fixed the handling of the array concern.

PR #188, opened 2017-08-12
Count oldLines and newLines when there are conflicts

haitaoli:
## Summary
Try to count `oldLines` and `newLines` when the hunk contains conflicts. If the patches have the same base file then `oldLines` should be valid. If base files are different then `oldLines` is deleted. `newLines` isn't very useful but I'm keeping the logic symmetric.


.gitignore
```diff
@@ -1,5 +1,3 @@
 coverage
 node_modules
-npm-debug.log
-lib
-dist
\ No newline at end of file
+npm-debug.log
```
kpdecker: These files are left out of the library by design to avoid the maintenance concerns they bring. Glad to take a look at a PR with just functional changes.

haitaoli:
Reverted .gitignore

kpdecker:
Can you remove the dist and lib files as well?

haitaoli:
Removed

PR #197, opened 2017-09-29
Fix: diffArrays can compare falsey items

### LINKED ISSUE #173: Bug: diffArrays gives wrong result on array of booleans
```
const jsdiff = require('diff')
const a1 = [true,true,true]
const a2 = [true,false,true]
const diff = jsdiff.diffArrays(a1, a2)
```

outputs

```
[ { count: 2, value: [ true, true ] },
  { count: 1, added: undefined, removed: true, value: [ true ] } ]
```

which is wrong! i'd expect something like

```
[ { count: 1, value: [ true ] },
  { count: 1, added: undefined, removed: true, value: [ true ] }
  { count: 1, added: true, removed: undefined, value: [ false ] }
  { count: 1, value: [ true ] } ]
```

pedrottimark:
Fixes #173 

* Override `removeEmpty` method
* Add a test that failed first

### DIFF
--- a/src/diff/array.js
+++ b/src/diff/array.js
@@ -4,5 +4,8 @@ export const arrayDiff = new Diff();
 arrayDiff.tokenize = arrayDiff.join = function(value) {
   return value.slice();
 };
+arrayDiff.removeEmpty = function(value) {
+  return value;
+};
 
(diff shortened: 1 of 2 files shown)


kpdecker:
Thanks!

PR #200, opened 2017-10-09
Added buffer diff configuration

woutervh-:
In my application I needed a way to apply diff to binary data.

My first attempt was to convert buffers to strings using `buffer.toString('binary')` but because the algorithm makes calls to `.join()` the resulting string was in UTF-8 encoding with extra characters for unicode characters outside ASCII range.

I ended up adding support for buffer directly so that changes properly contain the number of bytes in the `count` and the `value` is also a buffer which can then be converted to any of your favorite string encodings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,9 @@
   "scripts": {
     "test": "grunt"
   },
-  "dependencies": {},
+  "dependencies": {
+    "buffer": "^5.0.8"
+  },
   "devDependencies": {
     "async": "^1.4.2",
     "babel-core": "^6.0.0",
--- a/src/diff/buffer.js
+++ b/src/diff/buffer.js
@@ -1,0 +1,17 @@
+import {Buffer} from 'buffer';
+import Diff from './base';
+
+export const bufferDiff = new Diff();
+export function diffBytes(oldBuffer, newBuffer, options) { return bufferDiff.diff(oldBuffer, newBuffer, options); }
+
+/* Buffer itself can be accessed with array index syntax */
+bufferDiff.tokenize = (input) => input;
+bufferDiff.castInput = (input) => input;
+
+/* Simply create a buffer from the bytes */
+bufferDiff.join = (bytes) => {
+    return Buffer.from(bytes);
+};
+
+/* There is no notion of "empty" in buffer */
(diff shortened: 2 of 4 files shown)


woutervh-:
Side note: compatibility with browsers is achieved through the `buffer` package.

kpdecker:
I'm concerned about the weight of that lib for a use case that I don't envision people needing. Can you instead expose this functionality from diff/buffer?

woutervh-:
Just to be sure: it's the buffer npm package that you're worried about, right? In that case, we can just drop the import/export in index and leave the diff/buffer file as it. I'm just wondering what to do about the dependency in package.json?

ExplodingCabbage:
Hmm... it appears to me that `diffArrays` already works fine on `Buffer`s (and presumably on any type that provides indexing and a `slice` method?), making this unnecessary.

Even if that didn't work, I think I would still take issue with this:
* it seems weird and unintuive to use a generic name like `diffBytes` for a method that returns a type from a third-party library, when there are native JS ways of represent an array of bytes like [`Uint8Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array).
* I'd disagree with adding a dependency on a library just to be able to write a specialised diff for that library's types. Better for either that to live in its own package or for us to tweak jsdiff's API and documentation to make rolling your own diff for a new type (like you've done here) more trivial

With the above points in mind I don't think it makes sense to merge this PR and I don't think there are any possible changes to it that would change my mind, so I'm going to close it. I welcome argument if you disagree or see a flaw my reasoning, though!

ExplodingCabbage:
Oh, I forgot to paste an example of `diffArrays` working fine on a Buffer. Here it is:

```
> buffer.Buffer.from("abc")
<Buffer 61 62 63>
> x = buffer.Buffer.from("abc")
<Buffer 61 62 63>
> y = buffer.Buffer.from("abde")
<Buffer 61 62 64 65>
> diff.diffArrays(x,y)
[
  { count: 2, value: <Buffer 61 62> },
  { count: 1, added: undefined, removed: true, value: <Buffer 63> },
  { count: 2, added: true, removed: undefined, value: <Buffer 64 65> }
]
```

Issue #211, opened 2018-01-05
Should diffWordsWithSpace treat consecutive new lines as a single "word"?


attaboy:
The behavior of diffWordsWithSpace surprised me with its treatment of consecutive new lines as a single word.

For example:

Left:
```
A

B
```

Right:
```
A
B
```

Given the above, I would expect this diff:

`[{ value: "A\n" }, { value: "\n", removed: true }, { value: "B\n" }]`

Instead, I get:

`[{ value: "A" }, { value: "\n\n", removed: true}, { value: "\n", added: true}, { value: "B\n" }]`

Is this the expected behavior? I would love if it was configurable, because it's A) surprising and, B) if trying to display the diffs visually, it makes changes to new lines seem more complicated than they really are.

(I can see about opening a pull request if a change to this behavior is welcome.)

Mingun:
Just note that this is duplicate of #180.

ExplodingCabbage:
Fixed by https://github.com/kpdecker/jsdiff/pull/217.

PR #217, opened 2018-03-10
Threat each newline as separate word in `diffWordsWithSpace`

### LINKED ISSUE #180: diffWordsWithSpace - a char-by-char diff for spaces?
```js
require('diff').convertChangesToXML(
  require('diff').diffWordsWithSpace('foo\nbar', 'foo\n\n\nbar')
)
```

Returns:

```js
'foo<del>\n</del><ins>\n\n\n</ins>bar'
```

Desired output is:

```js
'foo<ins>\n\n</ins>bar'
```

As far as I understand, `diffWordsWithSpace` currently treats any amount of spaces as a single token. Which makes sense for `diffWords` (where space isn't significant), but doesn't work well for `diffWordsWithSpace`.

Suggestion: maybe it should compare spaces between words on character-by-character basis?

Possible solution - change regexp in [tokenizer](https://github.com/kpdecker/jsdiff/blob/8f8e0f296a6935a1854d1f2c0ea7ddac7e89a0ae/src/diff/word.js#L35) like this:

```diff
<   let tokens = value.split(/(\s+|\b)/);
---
>   let tokens = this.options.ignoreWhitespace ?
>                value.split(/(\s+|\b)/) :
>                value.split(/(\s|\b)/);
```

Seems to work fine for me, but I'm not quite sure it's correct.

Mingun:
This fixes #180 & #211

### DIFF
--- a/src/diff/word.js
+++ b/src/diff/word.js
@@ -32,7 +32,8 @@ wordDiff.equals = function(left, right) {
   return left === right || (this.options.ignoreWhitespace && !reWhitespace.test(left) && !reWhitespace.test(right));
 };
 wordDiff.tokenize = function(value) {
-  let tokens = value.split(/(\s+|[()[\]{}'"]|\b)/);
+  // All whitespace symbols except newline group into one token, each newline - in separate token
+  let tokens = value.split(/([^\S\r\n]+|[()[\]{}'"\r\n]|\b)/);
 
   // Join the boundary splits that we do not consider to be boundaries. This is primarily the extended Latin character set.
   for (let i = 0; i < tokens.length - 1; i++) {
--- a/test/diff/word.js
+++ b/test/diff/word.js
@@ -209,6 +209,15 @@ describe('WordDiff', function() {
       expect(convertChangesToXML(diffResult)).to.equal('&quot;<ins>word</ins>&quot;');
     });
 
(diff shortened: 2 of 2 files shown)


Mingun:
@kpdecker any feedback?

h2grover:
@Mingun how soon this fix will be merged?

Mingun:
@kpdecker, can we make progress with this? Better late than never

kpdecker:
Sorry for the delay on this, Have very few cycles for this project these days.

Mingun:
Nice to see that you don't forgot the project

ExplodingCabbage:
IMO the behaviour this produces for `diffWords` is broken as a result of the extra tokens. Here's a diff using the code _before_ this change...

```
> wd.diffWords("foo bar", "foo\n\nbar")
[ { value: 'foo\n\nbar', count: 3 } ]
```

... and here's one with the code _after_ this change:

```
> wd.diffWords("foo bar", "foo\n\nbar")
[
  { count: 2, value: 'foo\n' },
  { count: 1, added: true, removed: undefined, value: '\n' },
  { count: 1, value: 'bar' }
]
```

We're now only *sort-of-ignoring* whitespace in `diffWords`.

(In fairness, we were already doing a bad job of ignoring whitespace in `diffWords`; if you changed the whitespace around some *punctuation* it would be treated as a change, before.)

PR #218, opened 2018-03-10
Make ()[]"' as word boundaries between each other

Mingun:
This produces more natural diffs in situations like `()` -> `(word)`


Mingun:
@kpdecker any feedback?

PR #232, opened 2018-09-24
jsdiff: Replaced phantomJS by chrome

ossdev07:
Added Karma-chrome dependency for aarch64

Signed-off-by: ossdev07 <ossdev@puresoftware.com>


ossdev07:
Hi,

Removed phantomJS dependency from package.json.
PhantomJS is unmaintained, Remove the dependency and use headless chrome instead.

-> Reason for using --no-sandbox flag:

For aarch64 chrome binary is not present, we are using chromium binary to launch the Chrome.
Security is one of the most important goals for Chromium, and --no-sandbox flag provides that.
When trying to use chromium as root user like in docker container then only the requirement of --no-sandbox flag is coming
Without the flag, chromium gives error as root user-
~~~
[28769:28769:0925/163702.505790:ERROR:zygote_host_impl_linux.cc(89)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
Thus, we need to give the --no-sandbox flag when we launch chrome with chromium binary.
~~~
-> Justification for change karma version

Updated the karma version to 2.0.0 because the earlier karma 0.13.11 was very old, though we can use the earlier version too.

ossdev07:
Ping @kpdecker 

Can you please have a look at this PR.
It will be great to see chrome headless support for jsdiff.

PR #244, opened 2019-01-08
Normalize doc/ex with `Diff` package.json naming from npmjs.com

### LINKED ISSUE #243: DOM instantiation fails
Thanks for having this project available. :)

On to business...In the `./node_modules/diff/dist/diff.js` there is some code of:

``` js
(function (global, factory) {
  typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
  typeof define === 'function' && define.amd ? define(['exports'], factory) :
  (global = global || self, factory(global.Diff = {}));
}(this, function (exports) { 'use strict';
```
This redefines the Object reference to be `Diff` instead of `JsDiff` with the snippet of `global.Diff = {}` for DOM usage as detailed in the [current README.md](https://github.com/kpdecker/jsdiff/blob/305a2b33da24c4a5cd1a547529ce06ecb00cb0a5/README.md#examples) under "Basic example in a web page".

Was this intended as one of the breaking changes with your semver major switch? e.g. are you wanting to change the DOCS to match the [npmjs.com package name](https://www.npmjs.com/package/diff) instead of the [GitHub repo name](https://github.com/kpdecker/jsdiff)?

Solutions I can think of are:
1. Fix the build system *(which is currently beyond my expertise at the moment since I've never done this particular one)*
2. Fix the README.md to this new norm. *(I can PR this if needed/wanted but will also do the node identifiers for symmetry)*
3. Something else?

... e.g. What was intended?

Thanks for the look.

Affected versions:
* *diff*@4.0.0
* *diff*@4.0.1

Martii:
NOTES:
* Option 2

Closes #243

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,25 +15,25 @@ npm install diff --save
 
 ## API
 
-* `JsDiff.diffChars(oldStr, newStr[, options])` - diffs two blocks of text, comparing character by character.
+* `Diff.diffChars(oldStr, newStr[, options])` - diffs two blocks of text, comparing character by character.
 
     Returns a list of change objects (See below).
 
     Options
     * `ignoreCase`: `true` to ignore casing difference. Defaults to `false`.
 
-* `JsDiff.diffWords(oldStr, newStr[, options])` - diffs two blocks of text, comparing word by word, ignoring whitespace.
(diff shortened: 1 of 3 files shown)


Martii:
If this wasn't your intention please close. Thanks

Martii:
Fun GH is having issues on auto-closing things.

PR #283, opened 2020-04-04
build for browser

disarticulate:


kpdecker:
This is handled by the grunt command. What is the use case for running this independently of that?

kpdecker:
Closing as housekeeping, but glad to discuss.

PR #293, opened 2020-08-07
Feat: Add `exports` map for Native ESM

### LINKED ISSUE #292: Native ESM
Hey there, trying to add support native ESM within `uvu`, and `diff` is the only blocker. 

You have `main/module` setup, but in Node versions that support ESM, they're picking up the CommonJS/`
main"` build, which nests everything behind a `default` key.

To fix this, all you need to do is add this to your `package.json` file:

```js
"export": {
  ".": {
    "import": "./lib/index.mjs",
    "require": "./lib/index.js"
  },
  "./package.json": "./package.json"
}
```

And then kick out an extra `lib/index.mjs` file with the same contents as `lib/index.es6.js`.

Check [this repo](https://github.com/TehShrike/uvu-diff-repro) for a simple reproduction. 
It has nothing to do with `uvu` by the way, it's just how I uncovered the problem.

lukeed:
Closes #292 

Used the "loose" variant so that this won't be accidentally breaking for users who are spelunking into the `dist` or `lib` directories for specific file(s). 

There's an issue with Node 13.0 through 13.6 that shipped early, now-deprecated versions of the `exports` map syntax. 
Any users on those versions _should have already_ abandoned them, since any odd major release (especially early versions of it) are always experimental.

The current `exports` map works in all 12.x and 14.x versions (and 13.7+). 
It's completely ignored by older versions of Node since they don't know what it is.

---

While completely unrelated, I also added the `unpkg` key so that developers can do this, if they want:

```html
<script src="https://unpkg.com/diff"></script>
```

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -8,10 +8,16 @@ export default [
     output: [
       {
         name: 'Diff',
-        file: pkg.browser,
-        format: 'umd'
+        format: 'umd',
+        file: pkg.browser
       },
-      { file: pkg.module, format: 'es' }
+      {
+        format: 'esm',
+        file: pkg.module
+      }, {
+        format: 'esm',
+        file: pkg['exports']['.'].import
+      }
     ],
     plugins: [
       babel({
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   "main": "./lib/index.js",
   "module": "./lib/index.es6.js",
   "browser": "./dist/diff.js",
+  "unpkg": "./dist/diff.js",
   "exports": {
     ".": {
       "import": "./lib/index.mjs",
--- a/package.json
+++ b/package.json
@@ -30,6 +30,14 @@
   "main": "./lib/index.js",
   "module": "./lib/index.es6.js",
   "browser": "./dist/diff.js",
+  "exports": {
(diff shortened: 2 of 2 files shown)


rollup.config.js
```diff
@@ -8,10 +8,16 @@ export default [
...
+        file: pkg.browser
       },
-      { file: pkg.module, format: 'es' }
+      {
+        format: 'esm',
+        file: pkg.module
+      }, {
+        format: 'esm',
+        file: pkg['exports']['.'].import
+      }
```
calebeby, line 20: @lukeed What is the advantage of having two separate ESM builds?

lukeed: Two things:

1) I'm primarily preserving `diff`'s existing output files, so that anyone who was import/requiring into the module (eg, `require('diff/lib/index.es6.js')`) does find a surprise.

2) I did not add `"type": "module"` (nor do I want to) since that'd be a breaking change. Without that, need `.mjs` extension for ESM to be activated.

calebeby: Thanks, that makes sense!

lukeed:
Hey thanks for merge :) Any idea when next release will happen? Thanks!

PR #304, opened 2020-11-18
Update link to paper in readme

pitizzzle:
The old link was broke.
However this one does not seem to be official, either.
Better than nothing?


pitizzzle:
Alright. The link that is already in use works again ✔.
Seems like the web server was just temporarily shut down.

PR #305, opened 2020-11-24
Update .travis.yml

nageshlop:
This PR adds CI support for the IBM Power Little Endian (ppc64le) architecture. The idea is to ensure that the builds on this architecture are continuously tested along with the Intel builds (amd64) as this is part of the ubuntu distro on that architecture as well and detecting (and fixing) any issues or failures early would help to ensure that we are always up to date.
The build and test results are available at the below location.
https://travis-ci.org/github/nageshlop/jsdiff/builds/745639024

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 sudo: false
+
 env:
   global:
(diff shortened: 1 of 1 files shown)


nageshlop:
Hi @kpdecker 
Could you please review this request.

ExplodingCabbage:
Given the nature of jsdiff (it has no dependencies and just operates on JavaScript strings and arrays, without doing anything fancy involving system calls, C/C++ addons, or use of the file system) it's pretty much inconceivable to me that jsdiff could ever encounter a situation where the tests pass on one architecture but fail on another. As such, running CI on multiple architectures seems like a waste to me and I'm inclined to say no to this.

PR #321, opened 2021-05-22
Fix online demo URL

Mogztter:
GitHub pages now uses github.io domain.

Since https://kpdecker.github.io/jsdiff redirects to http://incaseofstairs.com/jsdiff/ we could also use this URL.


waldyrious:
I'd also suggest moving the link to the top of the README, to make it more visible and more easily accessible :)

Mogztter:
Why not, let's hear what @kpdecker has to say.

PR #328, opened 2021-09-19
Improve word tokenization for non-Latin characters

jihunleekr:
`Diff.diffWords` is not working on non-Latin characters like Korean.

### DIFF
--- a/src/diff/word.js
+++ b/src/diff/word.js
@@ -1,25 +1,19 @@
 import Diff from './base';
 import {generateOptions} from '../util/params';
 
-// Based on https://en.wikipedia.org/wiki/Latin_script_in_Unicode
-//
-// Ranges and exceptions:
-// Latin-1 Supplement, 0080–00FF
-//  - U+00D7  × Multiplication sign
-//  - U+00F7  ÷ Division sign
-// Latin Extended-A, 0100–017F
-// Latin Extended-B, 0180–024F
-// IPA Extensions, 0250–02AF
-// Spacing Modifier Letters, 02B0–02FF
-//  - U+02C7  ˇ &#711;  Caron
-//  - U+02D8  ˘ &#728;  Breve
-//  - U+02D9  ˙ &#729;  Dot Above
-//  - U+02DA  ˚ &#730;  Ring Above
-//  - U+02DB  ˛ &#731;  Ogonek
-//  - U+02DC  ˜ &#732;  Small Tilde
-//  - U+02DD  ˝ &#733;  Double Acute Accent
-// Latin Extended Additional, 1E00–1EFF
-const extendedWordChars = /^[a-zA-Z\u{C0}-\u{FF}\u{D8}-\u{F6}\u{F8}-\u{2C6}\u{2C8}-\u{2D7}\u{2DE}-\u{2FF}\u{1E00}-\u{1EFF}]+$/u;
+const spaceChars = ' \f\t\v\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff';
+let charsCannotBecomeWord = '';
+charsCannotBecomeWord += '\n\r';
+charsCannotBecomeWord +=
+  '\u0000-\u002F\u003A-\u0040\u005B-\u0060\u007B-\u007E'; // Basic Latin
+  charsCannotBecomeWord += '\u00A0-\u00BF\u00D7\u00F7'; // Latin-1 Supplement
+  charsCannotBecomeWord += '\u02B9-\u02DD\u02E5-\u02FF'; // Spacing Modifier Letters
+  charsCannotBecomeWord += '\u0300-\u036F'; // Combining Diacritical Marks
+  charsCannotBecomeWord += '\u1000-\u1FAFF'; // Mahjong Tiles - Symbols and Pictographs Extended-A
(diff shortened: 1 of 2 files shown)


src/diff/word.js
```diff
@@ -1,25 +1,19 @@
...
-// Latin Extended Additional, 1E00–1EFF
-const extendedWordChars = /^[a-zA-Z\u{C0}-\u{FF}\u{D8}-\u{F6}\u{F8}-\u{2C6}\u{2C8}-\u{2D7}\u{2DE}-\u{2FF}\u{1E00}-\u{1EFF}]+$/u;
+const spaceChars = ' \f\t\v\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff';
+let charsCannotBecomeWord = '';
+charsCannotBecomeWord += '\n\r';
+charsCannotBecomeWord +=
+  '\u0000-\u002F\u003A-\u0040\u005B-\u0060\u007B-\u007E'; // Basic Latin
+  charsCannotBecomeWord += '\u00A0-\u00BF\u00D7\u00F7'; // Latin-1 Supplement
+  charsCannotBecomeWord += '\u02B9-\u02DD\u02E5-\u02FF'; // Spacing Modifier Letters
+  charsCannotBecomeWord += '\u0300-\u036F'; // Combining Diacritical Marks
```
ExplodingCabbage, line 11: Hmm... this can't be right. Treating combining diacritics as word breaks is definitely wrong for European languages that I am familiar with (and I would've _thought_ that it was wrong universally), since it leads to outcomes like this:

```
> a = 'festou\u0300'
'festoù'
> b = 'impube\u0300re'
'impubère'
> diff.diffWords(a, b)
[
  { count: 1, added: undefined, removed: true, value: 'festou' },
  { count: 1, added: true, removed: undefined, value: 'impube' },
  { count: 1, value: '̀' },
  { count: 1, added: true, removed: undefined, value: 're' }
]
```

That diff *ought* to just show the one word in the old text being deleted and a totally new word from the new text being added, but instead it shows us preserving the *code point for the grave accent* in the middle of the word and adding and deleting stuff around the accent. That's nonsense from the perspective of French or any other language I know with accents; the `è` (an e with a grave accent) in impubère is simply a letter in the word, just like the `b` and `r` next to it, and doesn't semantically represent a word break.

Do combining diacritics work differently in Korean (... if they even exist in Korean?) or some other language you're familiar with such that it makes sense to treat them as word breaks? I am afraid I don't speak any non-European languages and am a bit out of my depth! I can imagine in theory a language where accents get used as punctuation, but I have never encountered such a thing...

src/diff/word.js
```diff
@@ -32,21 +26,29 @@ wordDiff.equals = function(left, right) {
...
+    } else if (cannotBecomeWordRegExp.test(char)) {
+      tokens.push(char);
+      prevCharType = '';
+    } else {
+      if(prevCharType === 'word') {
+        tokens[tokens.length - 1] += char;
+      } else {
+        tokens.push(char);
+      }
+      prevCharType = 'word';
```
ExplodingCabbage, line 49: Why do we need logic changes here? It's not obvious to me why the fix for Korean text should involve anything more than just treating Korean letters as letters instead of word boundaries...

src/diff/word.js
```diff
@@ -32,21 +26,29 @@ wordDiff.equals = function(left, right) {
...
-      tokens[i] += tokens[i + 2];
-      tokens.splice(i + 1, 2);
-      i--;
+  const tokens = [];
+  let prevCharType = '';
+  for (let i = 0; i < value.length; i++) {
+    const char = value[i];
+    if (spaceRegExp.test(char)) {
+      if(prevCharType === 'space') {
+        tokens[tokens.length - 1] += ' ';
```
ExplodingCabbage, line 35: In the edge case where there's a long run of spaces in the text somewhere, this is going to take O(n^2) time where n is the number of consecutive spaces.

It also rewrites all space characters of any kind to an ordinary `' '`.

Both these aspects of the behaviour here seem wrong!

ExplodingCabbage:
If I understand right, the problem is that right now we treat all CJK characters as if they were punctuation marks / word breaks, and the fix here treats them as letters instead. But:

* the fix here also messes with combining diacritics in ways that seem to me to break existing working behaviour for languages with accents
* the fix also changes other aspects of the logic of `tokenize` beyond which characters are treated as letters vs word breaks, and I can't figure out why
* the fix doesn't help us with Japanese or Chinese since those languages don't use spaces (and need a fundamentally different tokenization algorithm like the one provided by `Intl.Segmenter`). Doesn't by itself make doing this a bad idea, but makes me wonder if we ought to be making a more radical change...

I'll come back to this in due course. Would love to get your input in the meantime, @jihunleekr, but I understand if in the 2 years since you opened this PR you've lost interest!

src/diff/word.js
jihunleekr: I hadn't considered such cases due to my limited experience with other languages. It seems like an aspect that requires further thought.

src/diff/word.js
jihunleekr: I agree. It seems like an excessive change.

src/diff/word.js
jihunleekr: It seems like this is also an excessive change. My intention was to separate characters that should exist as single characters into individual tokens. (For example: U+0021, U+0022, ...)

PR #331, opened 2021-10-20
chore: update license file

eddiemonge:
Use the standard BSD 3-Clause license

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,31 +1,29 @@
-Software License Agreement (BSD License)
+BSD 3-Clause License
 
 Copyright (c) 2009-2015, Kevin Decker <kpdecker@gmail.com>
-
 All rights reserved.
 
-Redistribution and use of this software in source and binary forms, with or without modification,
-are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


ExplodingCabbage:
A word diff:

image

ExplodingCabbage:
I'm comfortable concluding that none of the changes shown in the diff above can possibly affect the *meaning* of the license, so, even though I am not the copyright holder and don't have any right to relicense the library, I feel okay using using my collaborator powers to merge this without checking with Kevin.

Is the version you suggest using actually a more standard version of the license, though? Let's find out... Hmm... some googling suggests yes. Googling the first line of your version finds Wikipedia and the Open Source Initiative presenting it as the BSD 3-clause license. Googling the version on master finds Oracle and a load of other random corporations licensing stuff under it. I'm not sure if the version Kevin used here is an old version and the standard wording has been tweaked since, or if the version Kevin used was a fork someone made once and pointlessly tinkered with that caught on? Either way, it seems like this PR really does move us onto the current standard wording and therefore we might as well merge it (though I don't see why it matters at all).

Merging!

PR #332, opened 2021-10-25
Document diffJson() options

cincodenada:
I ran into this because `undefined` keys in the diff don't show up, so I needed to supply an undefined replacement

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,8 @@ npm install diff --save
     Returns a list of change objects (See below).
     
     Options
-    * `stringifyReplacer`: A custom replacer function. Operates similarly to the `replacer` parameter to [`JSON.stringify()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter), but must be a function. `undefined` will _not_ be replaced by this function; see the next option.
-    *  `undefinedReplacement`: A value to replace `undefined` with. By default it will not be replaced, and replacements for `undefined` from `stringifyReplacer` will not be used.
+    * `stringifyReplacer`: A custom replacer function. Operates similarly to the `replacer` parameter to [`JSON.stringify()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter), but must be a function.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,6 +56,10 @@ npm install diff --save
 * `Diff.diffJson(oldObj, newObj[, options])` - diffs two JSON objects, comparing the fields defined on each. The order of fields, etc does not matter in this comparison.
 
     Returns a list of change objects (See below).
+    
+    Options
+    * `stringifyReplacer`: A custom replacer function. Operates similarly to the `replacer` parameter to [`JSON.stringify()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter), but must be a function. `undefined` will _not_ be replaced by this function; see the next option.
```
ExplodingCabbage, line 61: > `undefined` will _not_ be replaced by this function

I don't think this is true. Looking at the code, it seems like the *only* place that `undefinedReplacement` gets used is in the default implementation of `stringifyReplacer`, and that if you provide both, then `undefinedReplacement` simply gets ignored. There doesn't seem to be any special logic preventing `stringifyReplacer` from running on an `undefined` value either:

```
> diff.diffJson(
...     {x: undefined},
...     {x: "bob"}
... )
[
  { count: 1, added: undefined, removed: true, value: '{}' },
  {
    count: 3,
    added: true,
    removed: undefined,
    value: '{\n  "x": "bob"\n}'
  }
]
> 
> diff.diffJson(
...     {x: undefined},
...     {x: "bob"},
...     {
...         undefinedReplacement: "bob",
...     }
... )
[ { value: '{\n  "x": "bob"\n}', count: 3 } ]
> 
> diff.diffJson(
...     {x: undefined},
...     {x: "bob"},
...     {
...         undefinedReplacement: "jim",
...         stringifyReplacer: (k, v) => k == "x" ? "bob" : v
...     }
... )
[ { value: '{\n  "x": "bob"\n}', count: 3 } ]
> 
> diff.diffJson(
...     {x: undefined},
...     {x: "bob"},
...     {
...         stringifyReplacer: (k, v) => k == "x" ? "bob" : v
...     }
... )
[ { value: '{\n  "x": "bob"\n}', count: 3 } ]
```

Maybe what you wrote here was true when you opened the PR - dunno, haven't checked - but it's wrong now.

Issue #275, opened 2020-01-07
between windows text(CRLF) and Mac/Linux(LF) text.


cctakaso:
We find to a problem at diff between windows text(CRLF) and Mac/Linux(LF) text.
ex)
widnows text file
1. abc(CRLF)
2. def
------------
mac text file
1. abc(LF)
2. def

Now, this answer is as follow.
first line  is different !

line.js
7     linesAndNewlines = value.split(/(\n|\r\n)/);
-->
this is may must be ....
7     linesAndNewlines = value.split(/(\r\n|\n)/);

thanks for your support!

ExplodingCabbage:
Resolved by https://github.com/kpdecker/jsdiff/pull/344

PR #344, opened 2022-02-05
Option to strip trailing CR

### LINKED ISSUE #343: Option to strip/ignore cr at end of line(?)
Hello! 

I've recently created a RFC (npm/rfcs#522) for `npm diff` to have an option to ignore `cr` to allow better diffing between platforms.

The problem with `npm diff` is that if you're developing on a Windows machine and want to compare it to a package built on UNIX, or vice versa, every line will appear as changed since the lineendings are different.

In the RFC I mention [`git diff --ignore-cr-at-eol`](https://github.com/git/git/commit/e9282f02b2f21118f3383608718e38efc3d967e1) and GNU [`diff --strip-trailing-cr`](https://www.gnu.org/software/diffutils/manual/html_node/diff-Options.html) as prior art.

While investigating how this would be implemented, I notice that `diff` does not have an option for something like this. I tried to find any previous issues/PRs that talked about this but only really found #68 which seems more concerned with separating lines from lineendings (correct me if I'm wrong) and only mentions crlf in passing.

It does mention that `diffTrimmedLines()` should ignore line endings since it should trim the lines, but as far as I can tell, this is not the case currently (diff@5.0.0). I tested in a simple [example setup](https://github.com/oBusk/jsdiff-crlf). 

Would an option similar to those in git/GNU diff make sense in jsdiff too, or is this against the nature of this package?

In the previously mentioned issue, there is also a new method, which is now an option, the `newlineIsToken` option which treats line endings as separate tokens, so I suppose it would be possible to use `diffLines()` with `newLineIsToken: true` and then parse the diff to ignore changes to line ending tokens (CRLF and LF). There is no such support in the patch methods as far as I can tell.

(Probable duplicate: #275)

oBusk:
Simple idea/suggestion.

Fixes #343, #275

See #343 for in-depth explanation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ npm install diff --save
     Options
     * `ignoreWhitespace`: `true` to ignore leading and trailing whitespace. This is the same as `diffTrimmedLines`
     * `stripTrailingCr`: `true` to remove all trailing CR (`\r`) characters before perfoming the diff. 
-      This helps to get a useful diff when comparing files from UNIX/Windows respectively.
+      This helps to get a useful diff when diffing UNIX text files against Windows text files.
     * `newlineIsToken`: `true` to treat newline characters as separate tokens.  This allows for changes to the newline structure to occur independently of the line content and to be treated as such. In general this is the more human friendly form of `diffLines` and `diffLines` is better suited for patches and other computer friendly output.
 
     Returns a list of [change objects](#change-objects).
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ npm install diff --save
 
     Options
     * `ignoreWhitespace`: `true` to ignore leading and trailing whitespace. This is the same as `diffTrimmedLines`
+    * `stripTrailingCr`: `true` to remove all trailing CR (`\r`) characters before perfoming the diff. 
+      This helps to get a useful diff when comparing files from UNIX/Windows respectively.
     * `newlineIsToken`: `true` to treat newline characters as separate tokens.  This allows for changes to the newline structure to occur independently of the line content and to be treated as such. In general this is the more human friendly form of `diffLines` and `diffLines` is better suited for patches and other computer friendly output.
 
     Returns a list of [change objects](#change-objects).
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ npm install diff --save
       - `ignoreWhitespace`: `true` to ignore leading and trailing whitespace.
       - `newlineIsToken`: `true` to treat newline characters as separate tokens. This allows for changes to the newline structure to occur independently of the line content and to be treated as such. In general this is the more human friendly form of `diffLines` and `diffLines` is better suited for patches and other computer friendly output.
       - `stripTrailingCr`: `true` to remove all trailing CR (`\r`) characters before perfoming the diff. 
-        This helps to get a useful diff when comparing files from UNIX/Windows respectively.
+        This helps to get a useful diff when diffing UNIX text files against Windows text files.
 
 * `Diff.createPatch(fileName, oldStr, newStr, oldHeader, newHeader)` - creates a unified diff patch.
 
--- a/README.md
+++ b/README.md
@@ -83,6 +83,8 @@ npm install diff --save
       - `context` describes how many lines of context should be included.
       - `ignoreWhitespace`: `true` to ignore leading and trailing whitespace.
       - `newlineIsToken`: `true` to treat newline characters as separate tokens. This allows for changes to the newline structure to occur independently of the line content and to be treated as such. In general this is the more human friendly form of `diffLines` and `diffLines` is better suited for patches and other computer friendly output.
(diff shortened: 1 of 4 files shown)


src/diff/line.js
```diff
@@ -3,6 +3,11 @@ import {generateOptions} from '../util/params';
 
 export const lineDiff = new Diff();
 lineDiff.tokenize = function(value) {
+  if(this.options.stripTrailingCr) {
+    // remove all CR (\r) characters from the input string
+    value = value.replace(/\r+\n/g, '\n');
```
ExplodingCabbage, line 8: The comment, code, and analogy to GNU `diff`'s `--strip-trailing-cr` option all disagree about the behaviour here:

* `--strip-trailing-cr` only removes at most ONE `\r` before an `\n`
* the code here removes a sequence of any number of consecutive `\r` characters before a `\n`
* the comment implies that all `\r` characters will be removed, even if they don't appear before a `\n`

We should tweak the behaviour to be consistent with the GNU diff option whose name we're copying, and we should make the comment accurately reflect the behaviour.

README.md
```diff
@@ -37,6 +37,8 @@ npm install diff --save
 
     Options
     * `ignoreWhitespace`: `true` to ignore leading and trailing whitespace. This is the same as `diffTrimmedLines`
+    * `stripTrailingCr`: `true` to remove all trailing CR (`\r`) characters before perfoming the diff. 
+      This helps to get a useful diff when comparing files from UNIX/Windows respectively.
```
ExplodingCabbage, line 41: English nit: "respectively" here doesn't make sense.

Suggested wording:

> This helps to get a useful diff when diffing UNIX text files against Windows text files.

ExplodingCabbage:
I am broadly in favour of this change; it's useful and consistent with the precedent provided by `diff`. Just have some nits. I'll polish it up in due course if @oBusk doesn't get to it first!

src/diff/line.js
```diff
@@ -3,6 +3,11 @@ import {generateOptions} from '../util/params';
...
+  }
+
   let retLines = [],
       linesAndNewlines = value.split(/(\n|\r\n)/);
```
ExplodingCabbage, line 12: This should probably just be `value.split("\n")` after this change, right?

src/diff/line.js
oBusk: That's a good point! Have an update!

src/diff/line.js
oBusk: Well only if `stripTrailingCr===true`, to not be breaking. But even in that case, I'm not sure I understand the benefit?

ExplodingCabbage: Ah, I was getting confused by https://github.com/kpdecker/jsdiff/issues/275#issue-546119016. After reading that issue, but not testing anything to confirm, I had wrongly believed that the `\n` appearing first in the pipe-separated list of alternatives `\n|\r\n/` meant that it would always match in preference to `\r\n`, making the entire regex equivalent to just `/(\n)/`, and that the logic using that regex had thus always been fundamentally broken. I thus *thought* that with my comment above I was just proposing cleaning up some misleading code that never really worked.

But some quick experimentation suggests that I'm wrong, at least in Node, Chrome, and Firefox!

```
> "foo\r\nbar\r\nbaz\r\n".split(/(\n)/)
[
  'foo\r', '\n',
  'bar\r', '\n',
  'baz\r', '\n',
  ''
]
```

I'm not sure, then, why @cctakaso thought, in https://github.com/kpdecker/jsdiff/issues/275, that reordering the `\r\n` and `\n` in the regex would fix anything. Is it possible there is some JavaScript environment out there, that @cctakaso was using, with a regex engine where the order of alternatives really does affect the result in the way that issue suggests? I would've thought this would be standardised and all implementations would follow the standard, but tbh I don't want to spend hours unpicking the meaning of the ECMAScript 3 standard where regexes got added to JavaScript to confirm that the behaviour here has always been unambiguously specified!

My new suggestion, then, is that I'd be in favour of reversing the order as @cctakaso suggested, to `/(\r\n|\n)/`, just to make absolutely sure there's no ambiguity and avoid any need to research this - even though I'm now like 85% sure the order doesn't matter and this won't actually have any effect!

ExplodingCabbage: Oops, pasted the wrong thing. I obviously meant to paste this:

```
> "foo\r\nbar\r\nbaz\r\n".split(/(\n|\r\n)/)
[
  'foo', '\r\n',
  'bar', '\r\n',
  'baz', '\r\n',
  ''
]
```

oBusk: Yeah the problem #275 presents will be fixed by the option, but the solution they suggested wouldn't change much since the array that the regex produces would still contain all new lines as separate elements, and `'\n' !== '\r\n'`. 
I think there's just a misunderstanding and we shouldn't change the code for it.

ExplodingCabbage (approved):
LGTM. Thank you for the contribution and for coming back to tweak & comment & set me straight through multiple rounds of review! :)

PR #345, opened 2022-02-05
Document and test options to createPatch

oBusk:
The documentation states that createPatch only supports `context`, when in actuality it passes the options on to the diffLines implementation. 

This PR updates the documentation to clarify this and also adds tests to verify this behavior. 

Since these methods are supposed to output a patch, it's arguable that the options doesn't make sense, but I'd argue that the `ignoreWhitespace` at least makes sense since outputting a patch is a pretty normalized way to display a diff and it makes sense to be able to use the existing options for it. 

`newLineIsToken` is a bit weirder when creating "patches". I added docs and tests for that in separate [commit](https://github.com/kpdecker/jsdiff/commit/12eb1729c64ebcd3b9f1d81c289c9502d465bae9) so that it can be reverted if it's something that is not wanted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,9 @@ npm install diff --save
     * `newStr` : New string value
     * `oldHeader` : Additional information to include in the old file header
     * `newHeader` : Additional information to include in the new file header
-    * `options` : An object with options. Currently, only `context` is supported and describes how many lines of context should be included.
+    * `options` : An object with options. 
+      - `context` describes how many lines of context should be included.
+      - `ignoreWhitespace`: `true` to ignore leading and trailing whitespace.
 
 * `Diff.createPatch(fileName, oldStr, newStr, oldHeader, newHeader)` - creates a unified diff patch.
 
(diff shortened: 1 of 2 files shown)


kpdecker:
LGTM, thank you for the contribution.

ExplodingCabbage:
> `newLineIsToken` is a bit weirder when creating "patches". I added docs and tests for that in separate [commit](https://github.com/kpdecker/jsdiff/commit/12eb1729c64ebcd3b9f1d81c289c9502d465bae9) so that it can be reverted if it's something that is not wanted.

Yeah, IMO allowing `newLineIsToken` to be passed to patch functions at all is a misfeature. When there are actual changes to newline structure, the "patches" we get - like the one shown in the `newlineIsToken: true` test case added here - aren't valid patches and can't be applied to the original file by jsdiff or any other tool. I find it difficult to imagine this ever being useful to anyone but easy to imagine people getting caught out by it when they generate a "patch" with this option and then discover it can't be applied.

ExplodingCabbage:
However, let me also thank you for the tests - without them, I would never have realised this feature existed or that I was unwittingly breaking it in my third commit at https://github.com/kpdecker/jsdiff/pull/535! I think I want to the feature, but now I can at least be more deliberate about doing so and document it in the release notes...

PR #365, opened 2022-04-28
Allow early termination to limit execution time with degenerate cases

### LINKED ISSUE #364: Endless loop + hang with 100% CPU usage in execEditLength() when calling structuredPatch()
I encountered a hang in Yarn 3.2.0 and have isolated the root cause as an infinite loop in `diff@4.0.1`. (I'm not sure it's strictly infinite -- but it certainly doesn't terminate after an hour.) When Yarn asks `diff` to create a structured patch in order to show the user how the lockfile has changed, `diff` runs forever, consuming a lot of CPU while it does so.

I have reproduced the issue in [this gist](https://gist.github.com/xeger/f0af6d60a42892fa66562841577e693b) for your convenience.

I also verified that the hang still happens with `diff@5.0.0` -- so it's still a legitimate issue. If I make any progress on diagnosing or fixing, I will update this ticket.

xeger:
Fixes #364 by giving callers an option.

### DIFF
--- a/src/diff/base.js
+++ b/src/diff/base.js
@@ -29,7 +29,7 @@ Diff.prototype = {
 
     let newLen = newString.length, oldLen = oldString.length;
     let editLength = 1;
-    let maxEditLength = newLen + oldLen;
+    let maxEditLength = options.maxEditLength || newLen + oldLen;
     let bestPath = [{ newPos: -1, components: [] }];
 
     // Seed editLength = 0, i.e. the content starts with the same values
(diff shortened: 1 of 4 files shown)


src/diff/base.js
```diff
@@ -29,7 +29,7 @@ Diff.prototype = {
 
     let newLen = newString.length, oldLen = oldString.length;
     let editLength = 1;
-    let maxEditLength = newLen + oldLen;
+    let maxEditLength = options.maxEditLength || newLen + oldLen;
```
kpdecker, line 32: Can we cap this parameter at newLen+oldLen?

kpdecker:
LGTM outside of that one change.

xeger:
Done!

PR #384, opened 2022-11-02
Add `assumeLineDelimiter` to `applyPatch()` options

### LINKED ISSUE #157: Line Delimiters Missing on Structured Patch
```
let one = 'boop boop bleh\nasdf';
let other = 'bp boob blah\nfdsa';

let patch = jsdiff.createPatch('filename', one, other, 'oldheader', 'newheader');
let parsedPatch = jsdiff.parsePatch(patch);
let structuredPatch = jsdiff.structuredPatch('oldname', 'newname', one, other, 'oldheader', 'newheader');

// works
let patchedOther =  jsdiff.applyPatch(one, patch);

// works
let parsedPatchedOther = jsdiff.applyPatch(one, parsedPatch);

// Missing linedelimiters
let structuredPatchedOther = jsdiff.applyPatch(one, structuredPatch);
```

Throws an error on apply.js:108:43

### LINKED ISSUE #228: Applying a patch doesn't seem to work?
```javascript
[const jsdiff = require("diff");

let text = `# Markdown
This is some markdown.

Let's find the diffs
`;

let newText = `# Markdown
This is some markdown.

Let's find the diffs and prove it all works.
`;
let diff = jsdiff.structuredPatch("old", "new", text, newText);
let newDoc = jsdiff.applyPatch(text, diff);
console.log(newDoc);
```

and what I get back is:

```
/home/nicferrier/jswork/sqlparsing/node_modules/diff/lib/patch/apply.js:111
          delimiter = _hunk.linedelimiters[j];
                                          ^

TypeError: Cannot read property '0' of undefined
    at Object.applyPatch (/home/nicferrier/jswork/sqlparsing/node_modules/diff/lib/patch/apply.js:111:43)
```

I couldn't find a test that does this... but since structuredPatch is at the heart of all the patch routines that seems a bit odd.

Help?

### LINKED ISSUE #346: A structured patch fails to apply with a JavaScript error
diff v5.0.0
Node v16.13.1

Trying to apply a patch generated by `diff.structuredPatch` fails with this error:
```
node_modules\diff\lib\patch\apply.js:151
          delimiter = _hunk.linedelimiters[j];
                                          ^

TypeError: Cannot read properties of undefined (reading '0')
```
Here's the test script:
```JavaScript
var diff = require('diff');
var patch = diff.structuredPatch('path/to/original', '/path/to/file',
                                '1\n2\n3\n4', '1\nx\ny\n4', 'old header', 'new header');
console.log(diff.applyPatch('1\n2\n3\n4', patch));
```
`diff.applyPatch`, on the other hand succeeds, so this code runs fine:
```JavaScript
var diff = require('diff');
var patch = diff.createTwoFilesPatch('path/to/original', '/path/to/file',
                                '1\n2\n3\n4', '1\nx\ny\n4', 'old header', 'new header');
console.log(diff.applyPatch('1\n2\n3\n4', diff.parsePatch(patch)));
```
, and yields this output, as expected:
```
===================================================================
--- path/to/original    old header
+++ /path/to/file       new header
@@ -1,4 +1,4 @@
 1
-2
-3
+x
+y
 4
\ No newline at end of file

1
x
y
4
```
Looking at the structured patch generated by `diff.parsePatch`, it's different from the one produced by `diff.structuredPatch`, which is where the problem seems to be. Here's the former:
```
[
  {
    oldFileName: 'path/to/original',
    oldHeader: 'old header',
    newFileName: '/path/to/file',
    newHeader: 'new header',
    hunks: [
      {
        oldStart: 1,
        oldLines: 4,
        newStart: 1,
        newLines: 4,
        lines: [Array],
        linedelimiters: [Array]
      }
    ]
  }
]
```
Here's the latter:
```
{
  oldFileName: 'path/to/original',
  newFileName: '/path/to/file',
  oldHeader: 'old header',
  newHeader: 'new header',
  hunks: [
    {
      oldStart: 1,
      oldLines: 4,
      newStart: 1,
      newLines: 4,
      lines: [Array]
    }
  ]
}
```
It's missing `linedelimiters` in its hunks, which is what the error suggests. I didn't look at the outer object vs. array. I imagine it's fine for multiple patches, though.

rhysd:
Fixes #157
Fixes #228
Fixes #346

Currently return value from `structuredPatch` cannot be applied with `applyPatch`. This is because `linedelimiter` is missing in the hunks of the return value.

This PR adds new option `assumeLineDelimiter` to `applyPatch`. By setting the option, patches can be applied even if the line delimiter information is missing. So `applyPatch` can work with `structuredPatch`.

```js
// Let's say \n is used for newline
const old = ...;
const new = ...;
const patch = structuredPatch('file1', 'file2', old, new, 'header1', header2');
const patched = applyPatch(old, patch, { assumeLineDelimiter: '\n' });
assert.equal(new, patched);
```

### DIFF
--- a/src/patch/apply.js
+++ b/src/patch/apply.js
@@ -24,6 +24,7 @@ export function applyPatch(source, uniDiff, options = {}) {
       fuzzFactor = options.fuzzFactor || 0,
       minLine = 0,
       offset = 0,
+      assumeLineDelimiter = options.assumeLineDelimiter,
 
       removeEOFNL,
       addEOFNL;
@@ -89,7 +90,7 @@ export function applyPatch(source, uniDiff, options = {}) {
       let line = hunk.lines[j],
           operation = (line.length > 0 ? line[0] : ' '),
           content = (line.length > 0 ? line.substr(1) : line),
-          delimiter = hunk.linedelimiters[j];
+          delimiter = assumeLineDelimiter || hunk.linedelimiters[j];
 
       if (operation === ' ') {
         toPos++;
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ npm install diff --save
 
     - `fuzzFactor`: Number of lines that are allowed to differ before rejecting a patch. Defaults to 0.
     - `compareLine(lineNumber, line, operation, patchContent)`: Callback used to compare to given lines to determine if they should be considered equal when patching. Defaults to strict equality but may be overridden to provide fuzzier comparison. Should return false if the lines should be rejected.
+    - `assumeLineDelimiter`: String for line delimiters. When no line delimiter is set in the given patch, this method assumes line delimiter is the string. This option is useful when applying patch from `structuredPatch`.
 
 * `Diff.applyPatches(patch, options)` - applies one or more patches.
 
--- a/test/patch/apply.js
+++ b/test/patch/apply.js
@@ -721,6 +721,29 @@ describe('patch/apply', function() {
       expect(applyPatch(oldtext, diffed)).to.equal(newtext);
     });
 
+    it('should apply patch without linedelimiter when assumeLineDelimiter option is set', function() {
+      expect(applyPatch(
+          'line1\n'
+          + 'line2\n',
+          {
+            oldFileName: 'file.txt',
+            newFileName: 'file.txt',
+            oldHeader: 'header',
+            newHeader: 'header',
+            hunks: [{
+              oldStart: 1,
+              newStart: 1,
+              oldLines: 2,
+              newLines: 1,
+              lines: ['-line2']
+            }]
+          },
+          {
+            assumeLineDelimiter: '\n'
+          }))
+        .to.equal(
+          'line1\n');
+    });
 
   });
 


src/patch/apply.js
```diff
@@ -89,7 +90,7 @@ export function applyPatch(source, uniDiff, options = {}) {
       let line = hunk.lines[j],
           operation = (line.length > 0 ? line[0] : ' '),
           content = (line.length > 0 ? line.substr(1) : line),
-          delimiter = hunk.linedelimiters[j];
+          delimiter = assumeLineDelimiter || hunk.linedelimiters[j];
```
ExplodingCabbage, line 93: Hmm... this should be the other way round - i.e. `hunk.linedelimiters[j] || assumeLineDelimiter` - shouldn't it? As currently written, the line delimiter from `options` will take precedence over one actually provided in the hunk.

ExplodingCabbage:
I haven't scrutinized the existing code too deeply yet but my first thought is that a better fix might be to set `linedelimiters` to `['\n']` in `structuredPatch`, similarly to how `parsePatch` does. It looks like those two functions are *attempting* to return results of the same type, but the latter is missing the entire `linedelimiters` property that the former includes; we should probably fix that, at which point _this_ PR becomes moot?

ExplodingCabbage:
Closing in favour of https://github.com/kpdecker/jsdiff/pull/393; I don't think there's a need for the assumed delimiter to be configurable. (Even if we generate a patch using `.structuredPatch` on a file with Windows line endings, it'll just treat the lines as ending with `\r` characters and thus be handled fine by code that is hard-coded to think line endings are always `\n` characters.)

PR #395, opened 2023-03-02
[fix] str.split can not handle surrogate pair, replaced with Array.from

luoxzhg:
example
```
> Array.from("\n𝐀")
[ '\n', '𝐀' ]
```
```
> "\n𝐀".split('')
[ '\n', '\ud835', '\udc00' ]
```

### DIFF
--- a/src/diff/base.js
+++ b/src/diff/base.js
@@ -167,7 +167,7 @@ Diff.prototype = {
     return value;
   },
   tokenize(value) {
-    return value.split('');
+    return Array.from(value);
   },
   join(chars) {
     return chars.join('');


ExplodingCabbage:
Hmm, interesting. I fundamentally agree with the proposed change; operating on Unicode code points instead of UTF-16 code units is the correct default behaviour (and frankly no code should operate on UTF-16 code units unless it's a library explicitly for dealing with UTF-16). However, I don't want to merge any change that modifies the results that jsdiff emits without adding unit tests, adding release notes, and doing a major version number bump. I also want to carefully audit the code line by line to make sure there's nowhere else where we're similarly treating strings as sequences of UTF-16 code units instead of Unicode code points.

I want to churn through some of the more straightforward-to-handle issues and PRs before doing the above - but I _do_ intend to return to this PR in due course!

ExplodingCabbage:
Adding docs and stuff over at https://github.com/kpdecker/jsdiff/pull/500

Issue #105, opened 2016-04-05
feat: `applyPatch` reversely


benjycui:
It seems that we cannot apply patch on newStr to calculate oldStr now. I read source code https://github.com/kpdecker/jsdiff/blob/master/src/patch/apply.js#L85 , and thought that we can add an option to do so.

If you accept this feature, I will create a pull request :-)

kpdecker:
I'm not sure I fully understand the feature, but if you want to write some tests to demonstrate it, go for it.

Generally I'm fine with adding contributed features as long as they don't add a lot of overhead relative to their general usefulness and have proper tests.

Luiz-N:
@kpdecker I think @benjycui might be talking about the ability to "reverse" or "revert" a patch. I have a similar use case where if I apply a series of patches to a document how can I then "revert" them in the order they were applied. Is this possible or do I need to manually create a "reversed patch" for every one applied?

ericbiewener:
Here's my method for reversing a patch. Happy to open a PR if this is the right approach, but I'm unsure if it would cover all use cases.

```tsx
const reversePatch = (patch: string) => {
  const parsedDiff = diff.parsePatch(patch)[0]

  const { oldFileName, newFileName, oldHeader, newHeader, hunks } = parsedDiff

  parsedDiff.oldFileName = newFileName
  parsedDiff.oldHeader = newHeader
  parsedDiff.newFileName = oldFileName
  parsedDiff.newHeader = oldHeader

  for (const hunk of hunks) {
    const { oldLines, oldStart, newLines, newStart, lines } = hunk
    hunk.oldLines = newLines
    hunk.oldStart = newStart
    hunk.newLines = oldLines
    hunk.newStart = oldStart

    hunk.lines = lines.map(l => {
      if (l.startsWith('-')) return `+${l.slice(1)}`
      if (l.startsWith('+')) return `-${l.slice(1)}`
      return l
    })
  }

  return parsedDiff
}
```

gh-andre:
I was looking for the same feature and found this issue. It would be very useful to be able to apply a patch in reverse, like `patch --reverse` does. This allows one to avoid having to store the very first original text and instead makes it possible to apply patches in reverse against the latest version.

For example, for a Wiki post, blog post, etc, I would store just the latest text version and all patches from edits and I could show limited history of patches going back from the latest version and when people want to see some older version, I would apply patches in reverse to get the full text for that version.

PR #450, opened 2023-12-29
Add function to reverse a patch

ExplodingCabbage:
Resolves https://github.com/kpdecker/jsdiff/issues/105

(Code based on @ericbiewener's contribution in that issue.)

### DIFF
--- a/test/patch/reverse.js
+++ b/test/patch/reverse.js
@@ -1,0 +1,97 @@
+import {applyPatch} from '../../lib/patch/apply';
+import {structuredPatch, formatPatch} from '../../lib/patch/create';
+import {reversePatch} from '../../lib/patch/reverse';
+import {parsePatch} from '../../lib/patch/parse';
+
+import {expect} from 'chai';
+
+describe('patch/reverse', function() {
+  describe('#reversePatch', function() {
+    it('should output a patch that is the inverse of the provided patch', function() {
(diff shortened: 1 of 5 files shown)


test/patch/reverse.js
```diff
@@ -0,0 +1,97 @@
...
+        '@@ -227,5 +225,3 @@\n' +
+        ' \n' +
+        " * jsdiff keeps track of the diff for each diagonal using a linked list of change objects for each diagonal, rather than the historical array of furthest-reaching D-paths on each diagonal contemplated on page 8 of Myers's paper.\n" +
+        ' * jsdiff skips considering diagonals where the furthest-reaching D-path would go off the edge of the edit graph. This dramatically reduces the time cost (from quadratic to linear) in cases where the new text just appends or truncates content at the end of the old text.\n' +
+        '-\n' +
+        '-bar\n' +
+        '\n' +
+        '===================================================================\n' +
+        '--- b/CONTRIBUTING.md\t\n' +
+        '+++ a/CONTRIBUTING.md\t\n' +
```
ExplodingCabbage, line 84: I think it's a pre-existing issue orthogonal to the main point of this PR, but the way the original header gets screwed up here is a bit distasteful. I should dig in and make an issue before merging this.

ExplodingCabbage: Issue: https://github.com/kpdecker/jsdiff/issues/454

PR #461, opened 2024-01-08
Add function glyphDiff for emoji and higher UTF8 character support

alexander-riss:
the character diff currently does not handle high utf8 like emoji as their length is not calculated correctly. instead of breaking the old characterDiff method we now have a glyphDiff method for this purpose as it may perform worse than the existing method.
this PR extends the test suite and passes with 100% coverage

### DIFF
--- a/src/diff/glyph.js
+++ b/src/diff/glyph.js
@@ -1,0 +1,15 @@
+import Diff from './base';
+
+export const glyphDiff = new Diff();
+glyphDiff.tokenize = function(value) {
+  return [...value];
+};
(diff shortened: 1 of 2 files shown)


ExplodingCabbage:
I figured I'd rather do a breaking release making this change in `diffChars` (change already done via https://github.com/kpdecker/jsdiff/pull/500). Diffing UTF-16 code units is a pretty arbitrary and non-useful thing to do and I think it's better to simply have a single `diffChars` method that does the useful thing.

Issue #472, opened 2024-01-10
Document (and throw an error over) the incompatibility between newlineIsToken and ignoreWhitespace


ExplodingCabbage:
See https://github.com/kpdecker/jsdiff/issues/381#issuecomment-1885396094

PR #486, opened 2024-02-14
Sort out behaviour of newlineIsToken and ignoreWhitespace

ExplodingCabbage:
Three fixes here:

* `ignoreWhitespace` now behaves the same as other `ignoreXxx` methods - i.e. changes how tokens are compared for equality but does NOT affect the `value`s that show up in the returned change objects. Resolves https://github.com/kpdecker/jsdiff/pull/219.
* Fix the incompatibility between `newlineIsToken` and `ignoreWhitespace`. Resolves https://github.com/kpdecker/jsdiff/issues/472, albeit not quite in the way I'd originally intended.
* I've undocumented `diffTrimmedLines`. I don't see a reason to have two ways (`diffTrimmedLines` & `ignoreWhitespace`) to do exactly the same thing, so I've purged one from the *documented* API. (I'll still keep the method itself around just for backwards compat, though.)

For me to do before merging:
- [ ] Sanity check this works at all
- [ ] Add release notes. Remember to call out the deprecation of `diffTrimmedLines`.
- [ ] Take another look at https://github.com/kpdecker/jsdiff/pull/219. (I didn't read the diff there before doing my own work on this - I'd actually forgotten @Mingun had actually put up a *PR* about this and not just an issue.)
- [ ] Update the failing tests
- [ ] Add a new test confirming compatibility between `newlineIsToken` and `ignoreWhitespace`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,18 +48,10 @@ Broadly, jsdiff's diff functions all take an old text and a new text and perform
 * `Diff.diffLines(oldStr, newStr[, options])` - diffs two blocks of text, treating each line as a token.
 
     Options
-    * `ignoreWhitespace`: `true` to strip all leading and trailing whitespace characters from each line before performing the diff. Defaults to `false`.
+    * `ignoreWhitespace`: `true` to ignore leading and trailing whitespace characters when checking if two lines are equal. Defaults to `false`.
     * `stripTrailingCr`: `true` to remove all trailing CR (`\r`) characters before performing the diff. Defaults to `false`.
       This helps to get a useful diff when diffing UNIX text files against Windows text files.
     * `newlineIsToken`: `true` to treat the newline character at the end of each line as its own token. This allows for changes to the newline structure to occur independently of the line content and to be treated as such. In general this is the more human friendly form of `diffLines`; the default behavior with this option turned off is better suited for patches and other computer friendly output. Defaults to `false`.
-
-    Returns a list of [change objects](#change-objects).
-
-* `Diff.diffTrimmedLines(oldStr, newStr[, options])` - diffs two blocks of text, comparing line by line, after stripping leading and trailing whitespace. Equivalent to calling `diffLines` with `ignoreWhitespace: true`.
-
-    Options
-    * `stripTrailingCr`: Same as in `diffLines`. Defaults to `false`.
-    * `newlineIsToken`: Same as in `diffLines`. Defaults to `false`.
 
     Returns a list of [change objects](#change-objects).
 
--- a/src/diff/line.js
+++ b/src/diff/line.js
@@ -23,9 +23,6 @@ lineDiff.tokenize = function(value) {
     if (i % 2 && !this.options.newlineIsToken) {
       retLines[retLines.length - 1] += line;
     } else {
-      if (this.options.ignoreWhitespace) {
-        line = line.trim();
-      }
       retLines.push(line);
     }
   }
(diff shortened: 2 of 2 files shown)


test/diff/line.js
```diff
@@ -138,16 +138,39 @@ describe('diff/line', function() {
...
+      expect(convertChangesToXML(diffResult)).to.equal('line\r\n<del>old value \r\n</del><ins>new value\r\n</ins>line');
     });
+
+    it('should be compatible with newlineIsToken', function() {
+      const diffResult = diffTrimmedLines(
+        'line1\nline2\n   \nline4\n \n',
+        'line1\nline2\n\n\nline4\n   \n',
+        {newlineIsToken: true}
+      );
+      expect(convertChangesToXML(diffResult)).to.equal('line1\nline2\n<del>   </del>\n<ins>\n</ins>line4\n   \n');
```
ExplodingCabbage, line 163: Hmmm... this made sense to me when I wrote it, but on reflection it seems wrong for `<del>   </del>` to show up here. After all, a completely empty line and a line with a load of spaces are meant to be equal under `ignoreWhitespace`, right?

ExplodingCabbage: On even further reflection: I think the problem here is that:
* for `newlineIsToken` mode to make sense at all, completely blank line content shouldn't be a token (otherwise it's *strictly* the case that every second token is a newline character, so the newline characters will *always* be preserved, and the intended purpose of `newlineIsToken` - to allow changes to newline structure to be reflected in diff results independent of line content changes - is undermined).
* but to ensure that joining the non-removed bits of the diff actually yields the new text, non-blank, whitespace-only lines DO need to get a token!

Combining the two requirements above, we end up forced to treat deleting all the whitespace from a whitespace-only line as a deletion in `newlineIsToken` + `ignoreWhitespace` mode. I think this is therefore the right thing to do (if we're going to allow those settings to be used together at all) and I should just note it as a curiosity in the docs.

test/diff/line.js
```diff
@@ -138,16 +138,39 @@ describe('diff/line', function() {
...
+    });
+
+    it(
+      'should not consider adding whitespace to an empty line an insertion ' +
+      'even in newlineIsToken mode where a token may be an empty string',
+      function() {
+        const diffResult = diffTrimmedLines('foo\n\nbar', 'foo\n \nbar', {newlineIsToken: true});
+        expect(convertChangesToXML(diffResult)).to.equal('foo\n \nbar');
+      }
+    );
```
ExplodingCabbage, line 173: Ah, crap, I meant to remove this test before merging. It doesn't pass (and I added docs noting that!)

Mingun:
Hi @ExplodingCabbage!
My intention was to get correct diff behavior that you usually expect from code diffs. When I implemented my fix in #219 I think, I got the expected results, that TortoiseGit and similar tools give you when you merge file in ignored whitespace mode. I probably didn't actually investigate that behavior is identical, but I remember that at least results was not surprising:
- removed lines keeps formatting from "old" chunk (there is anyway no "new" chunk)
- added lines keeps formatting from "new" chunk (there is anyway no "old" chunk)
- don't remember what was returned for changed lines

Anyway, I think, that your fix is simpler, but I suggest to take tests from my PR, because they:
- shows all possible scenarios of space and newline combinations
  https://github.com/kpdecker/jsdiff/pull/219/files#diff-f6880fa6bc33ccb261ba2f8f22c2c97c990b0cbc56e64085d88b76d08d1c8de2R67-R85
- shows one relatively real-life scenario
  https://github.com/kpdecker/jsdiff/pull/219/files#diff-f6880fa6bc33ccb261ba2f8f22c2c97c990b0cbc56e64085d88b76d08d1c8de2R88-R106

I think, that having them it is much better to understand what to expect from results.

Issue #492, opened 2024-02-16
diffJson throws on `Object.create(null)` in some cases


danbeam:
Minimal repro:

```ts
diffJson(Object.assign(Object.create(null), {a: 123}), {b: 456});
```

With an error like:

```sh
/.../node_modules/diff/src/diff/json.js:68
      if (obj.hasOwnProperty(key)) {
              ^
TypeError: obj.hasOwnProperty is not a function
```

As far as I can tell, some popular libraries export object with these setups (e.g. `eslint` and `express`).  I believe I have a simple fix.

PR #493, opened 2024-02-16
Support `Object.create(null)` in JSON diffing

danbeam:
Addresses https://github.com/kpdecker/jsdiff/issues/492

I was recently trying to use `diff` to compare output from a library that makes use of `Object.create(null)`.  This seems valid (if not encouraged) in some situations, but `diff` was throwing as the `typeof` this is `'object'` and it is `!== null` but `obj.hasOwnProperty` does not exist.

Instead, use `Object.prototype.hasOwnProperty.call(obj, key)` similar to the way that `Object.prototyope.toString` is used elsewhere in this library.

### DIFF
--- a/src/diff/json.js
+++ b/src/diff/json.js
@@ -1,8 +1,9 @@
 import Diff from './base';
 import {lineDiff} from './line';
 
-const objectPrototypeToString = Object.prototype.toString;
-
+const objectPrototype = Object.prototype;
+const objectPrototypeToString = objectPrototype.toString;
+const objectPrototypeHasOwnProperty = objectPrototype.hasOwnProperty;
 
 export const jsonDiff = new Diff();
 // Discriminate between two lines of pretty-printed, serialized JSON where one of them has a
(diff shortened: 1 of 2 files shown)


src/diff/json.js
```diff
@@ -1,8 +1,9 @@
 import Diff from './base';
 import {lineDiff} from './line';
 
-const objectPrototypeToString = Object.prototype.toString;
-
+const objectPrototype = Object.prototype;
+const objectPrototypeToString = objectPrototype.toString;
+const objectPrototypeHasOwnProperty = objectPrototype.hasOwnProperty;
```
ExplodingCabbage, line 6: Stylistic opinion: these variables (including the existing `objectPrototypeToString` one that was there before your PR) are pointless indirection and it'd be clearer and nicer to just do `Object.prototype.hasOwnProperty.call(...)` instead of `objectPrototypeHasOwnProperty.call(...)`.

ExplodingCabbage: (I went ahead and tweaked this.)

ExplodingCabbage (approved):
Good catch; much appreciated!

PR #496, opened 2024-02-21
Adding Types

leonardobsjr:
Adding TS types.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "node": ">=0.3.1"
   },
   "main": "./lib/index.js",
+  "types": "./lib/index.d.js",
   "module": "./lib/index.es6.js",
   "browser": "./dist/diff.js",
   "unpkg": "./dist/diff.js",
@@ -43,7 +44,7 @@
   },
   "scripts": {
     "clean": "rm -rf lib/ dist/",
-    "build:node": "yarn babel --out-dir lib  --source-maps=inline src",
+    "build:node": "tsc && yarn babel --out-dir lib  --source-maps=inline src",
     "test": "grunt"
   },
   "devDependencies": {
@@ -82,6 +83,7 @@
     "rollup": "^1.0.2",
     "rollup-plugin-babel": "^4.2.0",
     "semver": "^7.3.2",
+    "typescript": "^5.3.3",
     "webpack": "^4.28.3",
     "webpack-dev-server": "^3.1.14"
   },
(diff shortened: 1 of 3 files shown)


ehoogeveen-medweb:
Another way would be to add jsdoc comments for the exported functions - IIRC those get picked up by tsc and used in the generated declaration files.

ExplodingCabbage:
Ah, nice! That does indeed seem to work and I guess it'll do the trick.

leonardobsjr:
Yeah, I understand that this is not ideal @ExplodingCabbage - This is just a way of making sure that every jsdiff version has up-to-date method information to be used in typescript. The issue that motivated me was not being able to use a method that existed on the latest version of jsdiff because the type was not updated on https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/diff.

That being said, @ehoogeveen-medweb suggestion is far better than porting the whole stuff lol

ehoogeveen-medweb:
Well, I don't know about *better* as I do like TypeScript a lot, but it's certainly a lot less work ;)

ExplodingCabbage:
Okay, coming back to this after a break...

I think porting to TypeScript is probably actually no more work than adding JSDoc comments would be. One problem with adding JSDoc comments is that some TypeScript concepts don't *quite* have clean analogues in JSDoc. For instance, the `diffChars` / `diffWords` / etc functions have a bunch of common options, so it makes sense to define a `BaseDiffOptions` interface and then have `DiffCharsOptions extends BaseDiffOptions`, `DiffWordsOptions extends BaseDiffOptions`, etc, but JSDoc doesn't seem to cleanly support this. Defining a shared type that's used in multiple places and then importing it from other files seems to be a bit weird/tricky/verbose in JSDoc?

I'm gonna have a stab at actually porting everything to TypeScript and using Babel's preset-typescript to transpile to JavaScript (as seems to be the recommendation at https://www.typescriptlang.org/docs/handbook/babel-with-typescript.html if you need a build pipeline anyway); I'm *hoping* that besides actually sprinkling TypeScript types into all the function definitions, there's not gonna be much work involved in this beyond adding a few lines to the Babel config. Will see if that hope is proved accurate...

ehoogeveen-medweb:
By the way, one thing that might help in this process is https://arethetypeswrong.github.io or https://www.npmjs.com/package/@arethetypeswrong/cli (right now there are no types, but once there are, getting CJS and ESM to both work can be tricky).

ExplodingCabbage:
Closing in favour of https://github.com/kpdecker/jsdiff/pull/591

PR #500, opened 2024-03-08
Make diffChars diff Unicode code points instead of UTF-16 code units

ExplodingCabbage:
Forked from https://github.com/kpdecker/jsdiff/pull/395. (Sorry @luoxzhg, I couldn't figure out how to push to that PR for some reason so had to open a new one.)


ExplodingCabbage:
Thanks for the contribution, @luoxzhg!

PR #503, opened 2024-03-19
Use npm-check-updates to upgrade all dev deps I can upgrade without breaking the tests

ExplodingCabbage:
Specifically, I ran:

    npx npm-check-updates -u --doctor --doctorTest "bash -c 'yarn clean && yarn grunt'"


ExplodingCabbage:
Changes to output in `dist/` and `lib/` folders seems to be limited to changes in ordering, whitespace, and comment positioning, all of which seems fine.

PR #505, opened 2024-03-19
Upgrade some more dev dependencies

ExplodingCabbage:


ExplodingCabbage:
Awww, man. This affects the built files in ways I simply don't understand:

```
--- dist-master/diff.js	2024-03-19 17:54:44.215622691 +0000
+++ dist/diff.js	2024-03-19 17:55:19.022355585 +0000
@@ -1,8 +1,8 @@
 (function (global, factory) {
   typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
   typeof define === 'function' && define.amd ? define(['exports'], factory) :
-  (global = global || self, factory(global.Diff = {}));
-}(this, (function (exports) { 'use strict';
+  (global = typeof globalThis !== 'undefined' ? globalThis : global || self, factory(global.Diff = {}));
+})(this, (function (exports) { 'use strict';
 
   function Diff() {}
   Diff.prototype = {
@@ -1827,6 +1827,4 @@
   exports.reversePatch = reversePatch;
   exports.structuredPatch = structuredPatch;
 
-  Object.defineProperty(exports, '__esModule', { value: true });
-
-})));
+}));
```

I guess I need to get some understanding of JavaScript module systems to evaluate if this breaks stuff. Argh!

ExplodingCabbage:
I'm just gonna roll forward with this, keep updating stuff, and before doing a release, figure out how to test that jsdiff still works with all module systems.

PR #543, opened 2024-08-20
Release 6.0.0-beta

ExplodingCabbage:


silverwind:
FYI, This beta was published as dist-tag [`latest`](https://www.npmjs.com/package/diff?activeTab=versions) which means `npm install diff` will install the beta, which may be unexpected.

ExplodingCabbage:
Oh, man! Thanks for letting me know; it's my first time helping maintain an npm package and I hadn't really noticed tags were a thing nor realised they'd affect `npm install` behaviour. (I wrongly assumed it was semver-aware and wouldn't install a pre-release version unless explicitly asked to; I should've checked with a test package.)

silverwind:
It's a common misunderstanding with npm, I've seen it too many times. To fix, you could publish the v6 beta again using `npm publish --tag beta` and then publish another v5 release without the `--tag` argument.

Issue #553, opened 2024-09-03
Combinations of numbers and letters no longer considered a single token in v6


greysteil:
First up, thanks for all your work on jsdiff, and for the v6 release.

We have a bunch of tests around diffs, and I noticed an unexpected change:

On v5:

```
// Diff.diffWordsWithSpace("at least $2m", "at least $4m")
Array [
  Object {
    "count": 5,
    "value": "at least $",
  },
  Object {
    "added": undefined,
    "count": 1,
    "removed": true,
    "value": "2m",
  },
  Object {
    "added": true,
    "count": 1,
    "removed": undefined,
    "value": "4m",
  },
]
```

On v6:
```
// Diff.diffWordsWithSpace("at least $2m", "at least $4m")
Array [
  Object {
    "added": false,
    "count": 5,
    "removed": false,
    "value": "at least $",
  },
  Object {
    "added": false,
    "count": 1,
    "removed": true,
    "value": "2",
  },
  Object {
    "added": true,
    "count": 1,
    "removed": false,
    "value": "4",
  },
  Object {
    "added": false,
    "count": 1,
    "removed": false,
    "value": "m",
  },
]
```

Not sure if this is a deliberate change or not? I think where words and numbers are combined it's probably best to treat them as a single Word.

efstathiosntonas:
I believe this also breaks emojis in react-native TextInput.

```tsx
const changes = diffChars(originalText, changedText) as CharactersDiffChange[];
```

will result in `"��"` instead of `"✌️😒"`

downgrading to 5.2.0 everything is fine.

ExplodingCabbage:
> Not sure if this is a deliberate change or not? I think where words and numbers are combined it's probably best to treat them as a single Word.

It wasn't deliberate, and I agree with you. Damn, I was really hoping I could pull off that 6.0.0 release without introducing any new bugs. :(

ExplodingCabbage:
@efstathiosntonas what you describe (with `diffChars`) sounds like an entirely separate bug that I've introduced in the same release. Can you provide minimal instructions on how to reproduce it, either here or in a new issue?

efstathiosntonas:
@ExplodingCabbage will create an expo repro!

greysteil:
You're the best @ExplodingCabbage ❤️

efstathiosntonas:
> I'll try to get a new release out _today_ with both these bugs fixed.

@ExplodingCabbage do you still need a repro for `diffChars`?

[...]

PR #554, opened 2024-09-05
Fix diffWords treating numbers and underscores as not being word characters

ExplodingCabbage:
Fixes https://github.com/kpdecker/jsdiff/issues/553 (the original issue, not the orthogonal bug posted in a comment).

The bug was introduced in https://github.com/kpdecker/jsdiff/pull/494. Previously we were splitting into words on [`\b`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Word_boundary_assertion), which per the linked MDN docs considers the following things to be word characters:

> Letters (A–Z, a–z), numbers (0–9), and underscore (_).

Then I swapped to using the `extendedWordChars` regex (previously only used for stitching together some adjacent tokens into single tokens after the original splitting), but didn't ensure it actually contained all the things that `\b` considers to be word characters.

This should fix it!

### DIFF
--- a/src/diff/word.js
+++ b/src/diff/word.js
@@ -19,7 +19,7 @@ import { longestCommonPrefix, longestCommonSuffix, replacePrefix, replaceSuffix,
 //  - U+02DC  ˜ &#732;  Small Tilde
 //  - U+02DD  ˝ &#733;  Double Acute Accent
 // Latin Extended Additional, 1E00–1EFF
-const extendedWordChars = 'a-zA-Z\\u{C0}-\\u{FF}\\u{D8}-\\u{F6}\\u{F8}-\\u{2C6}\\u{2C8}-\\u{2D7}\\u{2DE}-\\u{2FF}\\u{1E00}-\\u{1EFF}';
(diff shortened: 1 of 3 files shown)


src/diff/word.js
```diff
@@ -19,7 +19,7 @@ import { longestCommonPrefix, longestCommonSuffix, replacePrefix, replaceSuffix,
 //  - U+02DC  ˜ &#732;  Small Tilde
 //  - U+02DD  ˝ &#733;  Double Acute Accent
 // Latin Extended Additional, 1E00–1EFF
-const extendedWordChars = 'a-zA-Z\\u{C0}-\\u{FF}\\u{D8}-\\u{F6}\\u{F8}-\\u{2C6}\\u{2C8}-\\u{2D7}\\u{2DE}-\\u{2FF}\\u{1E00}-\\u{1EFF}';
+const extendedWordChars = 'a-zA-Z0-9_\\u{C0}-\\u{FF}\\u{D8}-\\u{F6}\\u{F8}-\\u{2C6}\\u{2C8}-\\u{2D7}\\u{2DE}-\\u{2FF}\\u{1E00}-\\u{1EFF}';
```
ExplodingCabbage, line 22: TBH I am confused about the ranges given here and by the comment explaining them above. I am gonna play around and maybe add more tests. But that can happen subsequent to just getting this PR merged.

PR #559, opened 2024-09-13
Ensure the result can be scrolled independently from the source

codebymikey:
When using the patch view, the patch results can be relatively small compared the source text, so it'd be useful if the results could be viewed and scrolled independently of the source.

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,7 +21,7 @@
 	<tr>
 		<td contenteditable="true" id="a">restaurant</td>
 		<td contenteditable="true" id="b">aura</td>
-		<td><pre id="result"></pre></td>
+		<td id="result_wrapper"><pre id="result"></pre></td>
 	</tr>
 </table>
 
--- a/style.css
+++ b/style.css
@@ -47,6 +47,10 @@ ins {
 	text-decoration: none;
 }
 
+#result_wrapper {
+	position: fixed;
+	overflow: scroll;
+}
 #result {
 	white-space: pre-wrap;
 }


ExplodingCabbage:
I'm guessing you're on a Mac?

On Ubuntu (and I imagine also Windows) the styling here is broken; the scrollbar for the `result_wrapper` element appears off-screen. (I think this bug won't have exhibited on a Mac because - if I remember right from when I last used one - scrollbars on Mac are inset inside the content box of the scrollable element and do not increase the size of any box, whereas on every other OS they appear outside the content box and increase the size of the border box.)

Setting `box-sizing: border-box` (and tweaking `width` to `33.3333%` instead of `33%`, to avoid a gap on the right) should fix it. I'll do that.

I also suggest there's no reason to make the result section a special case in this regard; better to make ALL three panes individually scrollable, no? If I have a diff where one input has a long prefix that another input doesn't, I may want to scroll past it and line up the similar regions with each other. I'll make that change too!

ExplodingCabbage:
> we can always include [synchronized scrolling] as an option if the UI makes sense for it

I think I lean towards thinking this isn't worth even the minor extra UI complexity of having a checkbox and label on the page. Only a minority of users will have a use for this feature and only a minority of those will ever derive any benefit from it beyond saving a few seconds of additional scrolling.

> Also on the list of features I planned to add when I had some free time was the ability to persist the text and patch options in the localStorage.

I think I feel the same about this. A user who wants to persist stuff can paste it into their editor (or paste it into Notepad and save some .txt files on their desktop or whatever). I don't want to add new UI elements whose only benefit is saving a minority of users a few seconds of work per use, in a webapp that I imagine most users only use rarely. This kind of thing might make sense in an application that people sit in front of all day doing their job in, but in this kind of context, simplicity is IMO more valuable.

(Persisting by default and having it non-configurable avoids the UI complexity but is a privacy risk for people on shared computers, so I wouldn't want to do that either.)

(Also, I figure that *part* of the usefulness of the webapp is as a demo of the abilities of the actual JavaScript library to people who are unfamiliar with it and don't know the scope of what it can do; UI features that don't map to underlying capabilities of the library perhaps therefore have potential to be confusing.)

I admit neither of these considerations are massively strong arguments against these features, but I personally think the added value of them is minor enough that these fairly weak arguments against are enough to mean that implementing them would be a net negative. Happy to be argued with on that point (or be overruled by @kpdecker if he disagrees)!

Issue #577, opened 2025-01-08
Type for Intl.Segmenter


BirgitPohl:
I'm installing `yeoman` 

![diff type](https://github.com/user-attachments/assets/d17aaa46-26b5-46b9-be6f-5efa4f528650)

```json
{
  "name": "generator-person",
  "version": "1.0.0",
  "description": ghjj",
  "license": "ISC",
  "author": "me",
  "type": "commonjs",
  "keywords": [
    "yeoman-generator"
  ],
  "files": [
    "generators"
  ],
  "main": "generators/index.js",
  "scripts": {
    "test": "test",
    "build": "npm run clean && tsc",
    "link": "npm link",
    "clean": "rm -rf generators"
  },
  "dependencies": {
    "ts-node": "^10.9.2",
    "typescript": "^5.7.2",
    "yeoman-generator": "^5.10.0"
  },
  "devDependencies": {
    "@types/yeoman-generator": "^5.2.14"
  }
}
```

When I
```bash
npm i
npm run build
```

I get an error, that

```bash
node_modules/@types/diff/index.d.ts:43:26 - error TS2694: Namespace 'Intl' has no exported member 'Segmenter'.

43     intlSegmenter?: Intl.Segmenter | undefined;
```

I wanted to report that here and not in definitely typed, because since you have added the support for this in 2024, you might want to check, if Intl has updated something that requires refactoring.

Currently, it blocks me.

lsmith77:
any update here?

PR #351, opened 2022-02-14
Fix `exports` field in `package.json`

fisker:
Fixes #350

[A solution that postss used](https://github.com/postcss/postcss/commit/92155c73e105a36af5196a75bcc34ada377a711f) seems unnecessary, as I tested, use `"./*": "./*"` works on Node.js 12 too.

PR #591, opened 2025-02-28
TypeScript rewrite

ExplodingCabbage:
Resolves https://github.com/kpdecker/jsdiff/issues/303
Resolves https://github.com/kpdecker/jsdiff/issues/577

See also discussion at https://github.com/kpdecker/jsdiff/pull/583.

This turned out to be a kind of hellish amount of work, but I think I've cracked it. This should eliminate the need for users to install `@types/diff` and also fix various issues with the types currently on DefinitelyTyped, along with making sure that henceforth there'll be no possibility of type declarations being out of sync with the underlying library. (TypeScript also caught two actual bugs in the library during the course of the rewrite, fixed separately in https://github.com/kpdecker/jsdiff/pull/601 and https://github.com/kpdecker/jsdiff/pull/602.)


ExplodingCabbage:
I think this is ready to merge. But goddamn, it was hard. Everything about this process was like waaaaay harder than expected, but most especially
* actually crafting roughly-correct types given both the nuance of how different possible `options` affect function return types and also [the obstacle that TypeScript's buggy matching of calls to overload signatures meant that some equally-correct possibilities were mysteriously broken](https://github.com/kpdecker/jsdiff/issues/303#issuecomment-2761606867)
* figuring out how to 1. avoid breaking any previously-supported ways of importing stuff while also 2. getting the types working for all module systems and arethetypeswrong giving my work its blessing. In particular:
  * I expected the right way to do things would be able to generate an ESM build, a CJS build, and a shared type declarations build that would be used as the types for both the ESM and CJS builds... but this seems almost impossible to do in a way that isn't broken (or at least that doesn't make attw howl with outrage - I wasn't always convinced its complaints were genuine problems, but also wasn't confident enough to ignore them), so instead I have identical copies of the types inside the `libesm` and `libcjs` folders. Seems weird, but whatever.
  * it seems to be outright impossible to use a folder mapping without an asterisk in `exports` in a way that `attw` won't error out about, but [we need it to support old versions of Node](https://github.com/kpdecker/jsdiff/pull/351), so I just had to use `--entrypoints` to tell `attw` to ignore it.

Probably there were multiple other things I banged my head against a wall for hours on, too - it's been a long time since I started work on this and I've forgotten things.

I've tried to review my own code and I reckon I'm happy merging and shipping this, but I'm going to wait at least a week. I'd be very grateful if anyone with more TypeScript expertise than me would be willing to review my work. @isaachinman & @andrewbranch, you seem like obvious candidates though of course you are not under any obligation at all to spend time looking at this.

isaachinman:
Not sure I will have time to review a `+3,647 −1,480` PR, but if you can release a canary version, I can install and check for type correctness/type errors.

ExplodingCabbage:
Alright, I'm gonna merge, then do a canary release as @isaachinman suggests.

ExplodingCabbage:
Version 8.0.0-beta is ready to play with!

Issue #608, opened 2025-05-12
Improve tree shakability


fisker:
```js
echo export{convertChangesToDMP}from"diff" | npx esbuild --bundle --format=esm
```

Expect output to only include `convertChangesToDMP`, instead of 800+ LOC

fisker:
Sorry I didn't notice that files in `libesm/` are exported, I should import from the file directly instead of the index file.

Great work! Thanks!

ExplodingCabbage:
Yep, I put that feature in specifically to avoid breaking the imports in Prettier that I recalled imported from `lib/` (though apparently that got changed in Pretter in the meantime and I hadn't noticed).

I am curious why esbuild can't manage to notice that those 800 lines of extra code are unused and purge them from the output, though!

fisker:
>  I put that feature in specifically to avoid breaking the imports in Prettier that I recalled imported from lib/ (though 

Thanks!


> I am curious why esbuild can't manage to notice that those 800 lines of extra code are unused and purge them from the output

I guess that's because lines like  this

https://github.com/kpdecker/jsdiff/blob/a5cf0db7c8ab3c2b3dc0de2e4e3afd754ba544de/src/diff/sentence.ts#L10

were not marked as "pure", can also caused by tansformed classes, I didn't dig, just my guess.

fisker:
I wonder if  put `sideEffects: false` in package.json will make the extra code go away.

fisker:
As I tested, `sideEffects: false` doesn't work. But My guess is correct, if I do

![Image](https://github.com/user-attachments/assets/90fc3ff4-0f48-4c01-9991-8e6479721121)

to all the files in `libesm/`


, I got

```
>echo export{convertChangesToDMP}from"diff" | npx esbuild --bundle --format=esm
// node_modules/diff/libesm/diff/word.js
var extendedWordChars = "a-zA-Z0-9_\\u{C0}-\\u{FF}\\u{D8}-\\u{F6}\\u{F8}-\\u{2C6}\\u{2C8}-\\u{2D7}\\u{2DE}-\\u{2FF}\\u{1E00}-\\u{1EFF}";
var tokenizeIncludingWhitespace = /* @__PURE__ */ new RegExp("[".concat(extendedWordChars, "]+|\\s+|[^").concat(extendedWordChars, "]"), "ug");

// node_modules/diff/libesm/convert/dmp.js
function convertChangesToDMP(changes) {
  var ret = [];
  var change, operation;
[...]

PR #612, opened 2025-05-16
Improve tree shakability

fisker:
Ref https://github.com/kpdecker/jsdiff/issues/608

As I tested, we only need change one place to fix it.

```
>yarn generate-esm && echo export{convertChangesToDMP}from"./libesm/index.js" | npx esbuild --bundle --format=esm
yarn run v1.22.19
$ yarn tsc --module nodenext --outDir libesm && echo {"type": "module","sideEffects":false} > libesm/package.json
$ E:\oss\forks\jsdiff\node_modules\.bin\tsc --module nodenext --outDir libesm
Done in 2.77s.
// libesm/convert/dmp.js
function convertChangesToDMP(changes) {
  var ret = [];
  var change, operation;
  for (var i = 0; i < changes.length; i++) {
    change = changes[i];
    if (change.added) {
      operation = 1;
    } else if (change.removed) {
      operation = -1;
    } else {
      operation = 0;
    }
    ret.push([operation, change.value]);
  }
  return ret;
}
export {
  convertChangesToDMP
};
```


package.json
```diff
@@ -71,8 +71,8 @@
     "clean": "rm -rf libcsm/ libesm/ dist/ coverage/ .nyc_output/",
     "lint": "yarn eslint",
     "build": "yarn lint && yarn generate-esm && yarn generate-cjs && yarn check-types && yarn run-rollup && yarn run-uglify",
-    "generate-cjs": "yarn tsc --module commonjs --outDir libcjs && echo '{\"type\": \"commonjs\"}' > libcjs/package.json",
-    "generate-esm": "yarn tsc --module nodenext --outDir libesm && echo '{\"type\": \"module\"}' > libesm/package.json",
+    "generate-cjs": "yarn tsc --module commonjs --outDir libcjs && echo {\"type\": \"commonjs\"} > libcjs/package.json",
+    "generate-esm": "yarn tsc --module nodenext --outDir libesm && echo {\"type\": \"module\",\"sideEffects\":false} > libesm/package.json",
```
fisker, line 75: Removed single quotes since it doesn't work on Windows.

package.json
```diff
@@ -127,5 +127,6 @@
     "lines": 100,
     "functions": 100,
     "statements": 100
-  }
+  },
+  "sideEffects": false
```
fisker, line 131: No harm to have this in the root package.json.

package.json
fisker: Make sense also to add `--target es6`? Class should not be transformed in ESM.

fisker: I'm not sure if TypeScript allows use `.mjs` as extension, better use `.mjs` and remove the package.json.

package.json
ExplodingCabbage: Hmm. `dist/diff.js` and `dist/diff.min.js` *do* have side effects - they create the `Diff` global. *Probably* this doesn't matter; I think there's no way anybody could be consuming those files that would respect `sideEffects: false`, given that they're not even exported via our `exports` property in `package.json`. But I am not *100%* sure of that. If it's not serving a purpose, I am inclined not to add this line here, just in case. (Leaving out the one that doesn't do anything also makes it clear which property is actually having an effect, if we want to change it later.)

package.json
```diff
@@ -71,8 +71,8 @@
     "clean": "rm -rf libcsm/ libesm/ dist/ coverage/ .nyc_output/",
     "lint": "yarn eslint",
     "build": "yarn lint && yarn generate-esm && yarn generate-cjs && yarn check-types && yarn run-rollup && yarn run-uglify",
-    "generate-cjs": "yarn tsc --module commonjs --outDir libcjs && echo '{\"type\": \"commonjs\"}' > libcjs/package.json",
-    "generate-esm": "yarn tsc --module nodenext --outDir libesm && echo '{\"type\": \"module\"}' > libesm/package.json",
+    "generate-cjs": "yarn tsc --module commonjs --outDir libcjs && echo {\"type\": \"commonjs\"} > libcjs/package.json",
```
ExplodingCabbage, line 74: I am not *sure* if it'll serve a purpose, but I figure we should add `"sideEffects": false` here too? It's true and could *plausibly* be useful, since a quick google reveals that tree-shaking plugins for CommonJS *do* exist. (I have no idea whether they respect `"sideEffects": false`, but even if they don't today, they might do tomorrow.)

package.json
ExplodingCabbage: > I'm not sure if TypeScript allows use .mjs as extension, better use .mjs and remove the package.json.

I think it always matches the extension of the files being compiled (so the only way to get `.mjs` output is to name your source files `.mts`). At least, I haven't found any other way to tweak the extensions that `tsc` outputs.

Honestly, I found setting up a TypeScript build pipeline that was compatible with everything and made arethetypeswrong happy to be a total nightmare, but I see that at least some popular TypeScript libraries like [Inquirer.js](https://www.npmjs.com/package/inquirer) use the same pattern I settled on for jsdiff (having a CJS folder and an ESM folder, using `.js` extensions in both, having a `package.json` in each, and duplicating the type declarations across the two of them) so I *think* what I'm doing here must be idiomatic and non-awful!

ExplodingCabbage: > Removed single quotes since it doesn't work on Windows.

Huh, interesting - do you personally develop on Windows? It looks to me like your changes here break this for *non*-Windows machines, since on my machine a package.json command like `"generate-cjs": "echo {\"type\": \"commonjs\"} > file` actually just writes `{type: commonjs}` to `file` (without the double quotes). I don't know CMD syntax well enough to know off the top of my head how to straightforwardly write this as a sh/CMD polyglot.

Are there any widely-accepted norms about how to go about supporting Windows in package.json scripts? It seems to me it'd be simpler to only support one species of shell language; perhaps I should put something in the README advising Windows people to use Git-Bash or WSL? Or I could wrap all shell commands in a `sh -c` invocation to ensure that, if they're executed at all, it's by `sh` and not CMD (but that's still only partly useful to Windows users unless they've got `sh` on their PATH).

ExplodingCabbage: > Make sense also to add --target es6? Class should not be transformed in ESM.

Hmm. Does it do any harm to transform them? They can still be instantiated and can still be extended with `extends`; is there anything that gets broken by the transformation?

fisker: No to all your questions, but it's unnecessary, since ESM version can't run in es5 environment, but it's up to you.

package.json
fisker: I didn't add it since Prettier only use ESM version 😄

package.json
fisker: > Are there any widely-accepted norms about how to go about supporting Windows in package.json scripts? 

Prettier uses `node --eval "fs.writeFileSync()"`

ExplodingCabbage (approved):
Thank you!

fisker:
Just want let you know, Prettier is able to save ~5K and can `import {} from "diff"` directly.  https://github.com/prettier/prettier/pull/17506

👍

PR #626, opened 2025-07-28
fix: export ArrayChange type

bdkopen:
Adds `ArrayChange` to the type exports. This was originally exported as part of the `@types/diff` package ([code](https://www.npmjs.com/package/@types/diff/v/7.0.2?activeTab=code)), but isn't exported here. Including it will make the v8 type update simpler.

Alternatively, we could cut the type altogether since it hasn't been exported since the v8 release.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -47,6 +47,7 @@ import {convertChangesToXML} from './convert/xml.js';
 import type {
   ChangeObject,
   Change,
+  ArrayChange,
   DiffArraysOptionsAbortable,
   DiffArraysOptionsNonabortable,
   DiffCharsOptionsAbortable,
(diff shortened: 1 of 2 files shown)


ExplodingCabbage:
Huh, interesting. It's a bit odd that I included a type with the same name here but make the type slightly different (not parameterized like the old DefinitelyTyped version) and didn't export it; clearly what I did doesn't make sense and must've been a mistake.

I think I see *three* options here would seem to be:
- export it, as this PR does
- just remove it, as you suggest - right now it's neither exported nor used internally
- export it *and* make it parameterized (`ArrayChange<T>` instead of just `ArrayChange`) like the old version

Issue #630, opened 2025-09-25
Double newline+Intl.Segmenter throw "string doesn't start with prefix; this is a bug"


SonOfLilit:
Minimal repro:

```
diffWords('A\n\nX', 'B\n\nX', {intlSegmenter: new Intl.Segmenter(["en"], { granularity: "word" })});

Uncaught Error: string "\nX" doesn't start with prefix "\n\n"; this is a bug
    at replacePrefix (diff.js?v=404c0ee8:237:11)
    at dedupeWhitespaceInChangeObjects (diff.js?v=404c0ee8:416:23)
    at diff.js?v=404c0ee8:382:11
    at Array.forEach (<anonymous>)
    at WordDiff.postProcess (diff.js?v=404c0ee8:375:13)
    at done (diff.js?v=404c0ee8:22:20)
    at execEditLength (diff.js?v=404c0ee8:70:18)
    at WordDiff.diffWithOptionsObj (diff.js?v=404c0ee8:96:21)
    at WordDiff.diff (diff.js?v=404c0ee8:17:17)
    at diffWords (diff.js?v=404c0ee8:400:19)
```

Versions:

Google Chrome Version 140.0.7339.208 (Official Build) (arm64)

```
    "node_modules/diff": {
      "version": "8.0.2",
      "resolved": "https://registry.npmjs.org/diff/-/diff-8.0.2.tgz",
      "integrity": "sha512-sSuxWU5j5SR9QQji/o2qMvqRNYRDOcBTgsJ/DeCf4iSN4gW+gNMXM7wFIP+fdXZxoNiAnHUTGjCr+TSWXdRDKg==",
      "license": "BSD-3-Clause",
      "engines": {
        "node": ">=0.3.1"
      }
    },
```

SonOfLilit:
(There's no issue when it's a single newline)

PR #631, opened 2025-10-06
Fix diffWords crashing when used with an Intl.Segmenter on a text with consecutive newlines

ExplodingCabbage:
Fixes https://github.com/kpdecker/jsdiff/issues/630

Thanks @SonOfLilit for reporting this!


AboAlroos:
# A first-level heading
## A second-level heading
### A third-level heading

Issue #634, opened 2025-10-07
word.ts extendedWordChars seems to not capture everything in the comment above it


hchiam:
in

https://github.com/kpdecker/jsdiff/blob/master/src/diff/word.ts#L5-L23

it currently is

```js
const extendedWordChars = 'a-zA-Z0-9_\\u{C0}-\\u{FF}\\u{D8}-\\u{F6}\\u{F8}-\\u{2C6}\\u{2C8}-\\u{2D7}\\u{2DE}-\\u{2FF}\\u{1E00}-\\u{1EFF}';
```

unless i'm misunderstanding something (e.g. exceptions mentioned in the comment, or how the `extendedWordChars` variable is actually interpreted by JS syntax) it seems to be missing 0080-00BF when i break down the unicode part of the string and compare it to the comment (which doesn't seem to mention 00F8-0249):

\\u{C0}-\\u{FF}
\\u{D8}-\\u{F6}
\\u{F8}-\\u{2C6}
\\u{2C8}-\\u{2D7}
\\u{2DE}-\\u{2FF}
\\u{1E00}-\\u{1EFF}

| extendedWordChars | comment |
| --- | --- |
**_SEEMS TO BE MISSING 0080-00BF?_**	| // Latin-1 Supplement, 0080–00FF
\\u{C0}-\\u{FF}	| // Latin-1 Supplement, 0080–00FF
_intentionally excluded_ | //  - U+00D7  × Multiplication sign
\\u{D8}-\\u{F6}	| _seems redundant since covered by other range \u{C0}-\u{FF}_
_intentionally excluded_	| //  - U+00F7  ÷ Division sign
\\u{F8}-\\u{2C6}	| **_COMMENT DOESN'T MENTION 00F8-0249?_**
_included in other ranges_	| // Latin Extended-A, 0100–017F
_included in other ranges_	| // Latin Extended-B, 0180–024F
_intentionally excluded_	| //  - U+02C7  ˇ \&#711;  Caron
\\u{2C8}-\\u{2D7}	| // IPA Extensions, 0250–02AF
_intentionally excluded_	| //  - U+02D8  ˘ \&#728;  Breve
_intentionally excluded_	| //  - U+02D9  ˙ \&#729;  Dot Above
_intentionally excluded_	| //  - U+02DA  ˚ \&#730;  Ring Above
_intentionally excluded_	| //  - U+02DB  ˛ \&#731;  Ogonek
[...]

PR #635, opened 2025-10-08
Fix some bugs in the diffWords regex (and errors & ambiguities in the comment above it)

ExplodingCabbage:
Resolves https://github.com/kpdecker/jsdiff/issues/634

(There are almost certainly still characters that this regex handles unreasonably, but this PR is not intended to fix all such cases; it's instead meant to fix the inconsistency between the comment and the regex noticed by @hchiam. It also fixes the handling of soft hyphens just because it was natural to consider that when looking at this.)


ExplodingCabbage:
Okay, well that "review" was completely useless. Whatever. Merging.

Issue #624, opened 2025-07-24
Is there a way to remove the `======` separator that appears at the top of diffs?


iryan2:
I'm using `jsdiff` with `react-diff-view` and am finding myself having to remove the `===================================================================` line that [appears at the top](https://github.com/kpdecker/jsdiff/blob/329f85b09018f5cf55000a58786163561e037bd2/src/patch/create.ts#L266) of the output diff, like so:

```
    if (lines.length > 0 && lines[0].startsWith("======")) {
        // Find the first --- line and start from there
        const headerStart = lines.findIndex((line) => line.startsWith("---"));
        if (headerStart >= 0) {
            return lines.slice(headerStart).join("\n");
        }
    }
```

The reason appears to be the same as is [described in this issue](https://github.com/otakustay/react-diff-view/issues/107#issuecomment-742400024), in that `react-diff-view` expects diffs to be in either git diff or unidiff format.

Can you please help me understand why it gets included in `jsdiff` output, and whether there's a way to prevent it? I'd love to avoid having to strip it myself.

Thanks!

iryan2:
Closing since I discovered that `react-diff-view` is [expected to be used](https://github.com/otakustay/react-diff-view/issues/64#issuecomment-525665282) with `unidiff` (a `jsdiff` wrapper), not `jsdiff` directly.

ExplodingCabbage:
Hmm. There's some conceptual confusions here.

The patches jsdiff emits _are_ in unified diff format (aka unidiff format). So too are the diffs that Git normally outputs. (It's "combined" format for showing merge diffs, enabled with the `-c` or `--c` flag, is not.) So when @otakustay [says](https://github.com/otakustay/react-diff-view/issues/107#issuecomment-742400024) that react-diff-view supports two formats, Git's format and unidiff format, that's already a bit confusing/misleading.

Or at least, we *say* the patches jsdiff emits are in unified diff format. Unfortunately unified diff format does not have any rigorous written spec, so a notion of what constitutes a "correct" unified diff format patch has to be reverse-engineered from what existing tools do - especially the GNU `diff -u` and `patch -u` utils. But tools output different header information - see the writeup that the DiffX guys have at the start of [their manual](https://diffx.org/). Whether the line `===================================================================` gets included after each Index line (and indeed whether Index lines even get included at all) is one such inconsistency; there are others.

[...]

PR #641, opened 2025-12-31
Allow more customisation of file headers in patches

ExplodingCabbage:
Resolves the jsdiff side of https://github.com/kpdecker/jsdiff/issues/624 (although as I describe there I think there are probably related changes that are sensible for other tools to make).

### DIFF
--- a/src/patch/create.ts
+++ b/src/patch/create.ts
@@ -281,6 +281,13 @@ export function formatPatch(patch: StructuredPatch | StructuredPatch[], headerOp
     headerOptions = INCLUDE_HEADERS;
   }
   if (Array.isArray(patch)) {
+    if (patch.length > 1 && !headerOptions.includeFileHeaders) {
+      throw new Error(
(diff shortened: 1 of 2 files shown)


ExplodingCabbage:
The extra detail in the [Copilot review session page](https://github.com/copilot/tasks/pull/PR_kwDOABeIS867HBgF?session_id=74473298-6f8e-40f8-89a5-fd5d0137161a) shows it caught some mistakes of mine that it didn't actually manage to post on the PR, for some reason:

> The `HeaderOptions` interface should be exported as a type so TypeScript consumers can use it when calling `formatPatch` or specifying the `headerOptions` field in the options for `createPatch` and `createTwoFilesPatch`.

Yep, good catch.

> The documentation for `formatPatch` should be updated to include the new optional `headerOptions` parameter.

(also createPatch maybe?). Also a good catch. Not sure why this didn't get posted.

Issue #644, opened 2026-01-04
ReDoS vulnerability in diff package via crafted input to parsePatch function leads to Node.js DoS


ShiyuBanzhou:
## Summary: 
  A Regular Expression Denial of Service (ReDoS) vulnerability was identified in the diff package, as distributed within Node.js (via deps/npm/node_modules/diff). The vulnerability is present in the dist/diff.js component at lines 677 and 712. An attacker can provide specially crafted input to the affected diffing functions, triggering catastrophic backtracking in the regular expressions. This leads to excessive CPU consumption and extremely long execution times, resulting in a Denial of Service (DoS) for any Node.js application or CLI tool utilizing these specific diffing utilities.
## Description: 
  Node.js (up to commit 559985c) is vulnerable to Regular Expression Denial of Service (ReDoS) through its bundled npm dependency, specifically within the diff library. Inefficient regex patterns at lines 677 and 712 of deps/npm/node_modules/diff/dist/diff.js allow for resource exhaustion. Successful exploitation allows a remote or local attacker to cause 100% CPU utilization by submitting malicious strings designed to trigger catastrophic backtracking, effectively causing a denial of service.
Steps To Reproduce:
PoC.js to trigger the vulnerability.
```js
/**
 * ReDoS PoC for regexId: 12
 *
 * Target Regex: /^(?:Index:|diff(?: -r \w+)+)\s+(.+?)\s*$/
 * Dataflow Path: uniDiff -> split(/\r\n|[\n\v\f\r\x85]/) -> diffstr[i] -> regex.exec()
 * Path Constraints: Input must contain Index: or diff prefix
 * Data Transformations: split() removes line delimiters (\r\n, \n, \r, \v, \f, \x85)
 *
 * === Verification Results ===
 * Phase 2 Result: Failed (Sink time: 0.029ms)
 * Phase 3 Iterations: 1
 * Iteration 1: Replaced \n with \u2028 in suffix → SUCCESS
 * Final Sink Execution Time: 131,440 ms (2m 11s)
 * Total Execution Time: 131,441 ms
 * Verification Time: 2025-11-28
 * Final Status: ✓ Verified and Passed
 * Success Criteria: Standard A (>> 5000ms)
 *
 * Key Fix: Used \u2028 (line separator) instead of \n in suffix.
 * Reason: split() removes \n but not \u2028, and . cannot match \u2028
 *
 * Generation Time: 2025-11-28
 * Status: ✓ Production Ready
 */

[...]

PR #649, opened 2026-01-07
Fix the second denial-of-service vulnerability in parsePatch

ExplodingCabbage:
This fixes the second, uh, ReDOS reported in https://github.com/kpdecker/jsdiff/issues/644. Though there's worse here than just a ReDOS!

First, I will explain the ReDOS. The regex `/^(---|\+\+\+)\s+(.*)\r?$/` takes n^2 time to execute against a string that ends with n spaces followed by a \u2028 and a non-space character, for similar reasons to those explained in https://github.com/kpdecker/jsdiff/pull/647. The change here fixes that.

But it was worse than that! The bigger problem is that the code around this regex used to go into an infinite loop (consuming more and more memory until an OOM error occurs) when called with even small adversarial inputs. e.g.:

```
> const { parsePatch } = require('./libcjs');
undefined
> parsePatch('--- x\u2028x')

<--- Last few GCs --->

[87253:0x45b0d000]    37408 ms: Mark-Compact 4051.2 (4131.1) -> 4042.3 (4138.1) MB, pooled: 0 MB, 972.26 / 0.00 ms  (average mu = 0.160, current mu = 0.127) allocation failure; scavenge might not succeed
[87253:0x45b0d000]    38452 ms: Mark-Compact 4058.0 (4138.1) -> 4046.7 (4142.6) MB, pooled: 0 MB, 1037.85 / 0.00 ms  (average mu = 0.114, current mu = 0.006) allocation failure; scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 0xe36196 node::OOMErrorHandler(char const*, v8::OOMDetails const&) [node]
 2: 0x123f4a0 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, v8::OOMDetails const&) [node]
 3: 0x123f777 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, v8::OOMDetails const&) [node]
 4: 0x146d1a5  [node]
 5: 0x146d1d3  [node]
 6: 0x148628a  [node]
 7: 0x1489458  [node]
 8: 0x1cc6071  [node]
Aborted (core dumped)
```

This happens because the line `--- x\u2028x` triggers the break on this line...

```
if ((/^(Index:\s|diff\s|---\s|\+\+\+\s|===================================================================)/).test(line)) {
        break;
```

... but does NOT match the regex at the start of `parseFileHeader`:

```
const fileHeader = (/^(---|\+\+\+)\s+(.*)\r?$/).exec(diffstr[i]);
if (fileHeader) {
  ...
```

(The reason it doesn't match is again down to the nasty trap that the regex wildcard `.` does not match character `\u2028`.)

The logic for chomping through lines of the patch relies on those two regexes being consistent with each other; since they are not, `parseIndex` finishes execution without actually having chomped the line, and gets called again by the main `while` loop in `parsePatch`, resulting in the line `list.push(index);` near the start of `parseIndex` being called again, resulting in `list` growing larger on each call until an OOM eventually occurs.

Oops!

This PR fixes it.


SSP6904:
Is there going to be a fix for AstroJS? This vulnerability patch doesn't work with the newest version of the framework.
image

PR #656, opened 2026-01-23
Backport redos 3.5.1

mukilane:
@ExplodingCabbage I am trying to backport the redos fix to 3.x.x as well since it is being a transient dependency from other packages that haven't released in a long time. I checked npm downloads, and v3 still has 4 million weekly downloads, so figured we might as well patch it here. image

(Since there is no branches over v3.5.0, I have opened the PR against master)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "diff",
-  "version": "3.5.0",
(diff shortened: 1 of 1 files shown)


ExplodingCabbage:
Released - thanks.

PR #661, opened 2026-02-09
Upgrade dev deps to latest versions with `yarn upgrade --latest`

ExplodingCabbage:
None of this matters, probably. But it should close some Dependabot PRs, and doing it from time to time helps keep up with the times and avoid depending on old versions of things that nobody uses any more that may become hard to find proper docs on over time.


ExplodingCabbage:
My limited testing/review:
- `yarn && yarn lint && yarn test` still passes in a clean repo
- https://eslint.org/blog/2026/02/eslint-v10.0.0-released/ doesn't seem to contain any relevant changes besides dropping support for old Node versions
- If I introduce a linting error deliberately, `yarn lint` catches it

(I haven't even looked at the release notes for any other major version bumps; I am happy to assume they are unproblematic as long as `yarn test` passes.)

PR #701, opened 2026-08-16
fix: don't treat a literal CR at EOF as a Windows line ending

maximilliangrand:
## Bug

`applyPatch` corrupts output when a Unix source is patched with a patch whose final line's content ends in a literal carriage return and has no trailing newline. The `\r` is silently dropped.

```js
import { structuredPatch, applyPatch } from 'diff';
const oldFile = 'line1\nline2\n';
const newFile = 'line1\nline2\nline3\r';
const patch = structuredPatch('f', 'f', oldFile, newFile, undefined, undefined, {context: 0});
applyPatch(oldFile, patch); // => 'line1\nline2\nline3'  (the '\r' is lost)
```

I found this by fuzzing random source/target pairs through `structuredPatch` -> `applyPatch`; 80 of 200k pairs failed to round-trip, all of this shape.

## Cause

`isWin` and `winToUnix` treat any patch line ending in `\r` as a Windows CRLF line ending. But a line ending in `\r` that is immediately followed by a `\ No newline at end of file` marker is not a CRLF ending (Windows endings are `\r\n`) - it's a literal carriage return in the final line's content. A genuine Windows no-newline-at-EOF line never ends in `\r`.

So `isWin` misclassifies such a patch as Windows, and when `applyPatch` (with `autoConvertLineEndings` on) converts it to match the Unix source, `winToUnix` strips the `\r`.

`unixToWin` already guards this exact case (`hunk.lines[i + 1]?.startsWith('\\')`), so it won't add a `\r` to a no-newline final line. This change adds the symmetric guard to `isWin` and `winToUnix`.

## Fix

Ignore a trailing `\r` that sits immediately before a `\ No newline at end of file` marker when deciding whether a patch is Windows / when stripping CRs.

Added tests: a `winToUnix` unit test, an `isWin` unit test, and an `applyPatch` round-trip regression test. Full suite passes with 100% coverage.

### DIFF
--- a/release-notes.md
+++ b/release-notes.md
@@ -4,6 +4,7 @@
 
 - [#697](https://github.com/kpdecker/jsdiff/pull/697) *`diffJson` now correctly handles JSON objects containing a key named `__proto__`*. (Previously, the returned diff would be as if the `__proto__` key did not exist on either of the objects being diffed.)
 - [#700](https://github.com/kpdecker/jsdiff/pull/700) *`diffJson` now correctly handles JSON objects containing a non-callable property named `toJSON`* - i.e. it gives such a property no special behaviour whatsoever, just as `JSON.stringify` doesn't. Previously, such properties caused an error to be thrown. (*Callable* `toJSON` properties continue to get the same special behaviour that `JSON.stringify` gives them.)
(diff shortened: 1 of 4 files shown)


ExplodingCabbage:
> when a Unix source is patched with a patch whose final line's content ends in a literal carriage return and has no trailing newline

It's more specific than that, no? Roughly, the patch also needs to not feature any other lines, in order for `isWin` to return `true` - i.e. it needs to be a single-line patch. Still a legit bug, though!

ExplodingCabbage (approved):
LGTM, thank you! Will just see if Copilot offers any useful input before I merge.

PR #707, opened 2026-09-19
Fix diffJson serialization of arrays with toJSON

jakezwang:
Apply `toJSON()` before deciding whether to canonicalize an array or an object. Previously, an array returned by `toJSON()` became an object with numeric keys, and an array's own `toJSON()` was ignored.

Add regressions for both cases, including sorted object keys inside a returned array.

Validation: `yarn test` passes on Node 22 and 24 (320 tests, build, lint, and type checks); Firefox browser tests pass (312 tests).


ExplodingCabbage:
Well spotted and thanks for the fix. Looks like this bug has been around since `toJSON` support was first added a decade ago in https://github.com/kpdecker/jsdiff/commit/23163b2f6742d281a3e42530dca7150c934f0a16; I think Kevin was purely thinking of it as supporting serialization of `Date`s (the only built-in type in JavaScript with a `toJSON` function at that time) and not thinking about the implications for user-provided `toJSON` functions.

Will add release notes and merge.

